Issue #16, opened 2023-01-13
Admin endpoint for setting user tiers


WyseNynja:
There is currently a helper script for this (`web3_proxy_cli change_user_tier_by_address`), but an admin endpoint would make this faster for our users since our admins would not need the infra team to run this command for them.

- [ ] admin endpoint
- [ ] when the tier is changed, clear the cache

yenicelik:
This was merged (and tested in prod as I understand), should be ready to be closed

PR #17, opened 2023-01-17
Admin endpoint to modify user tier

yenicelik:
#16 

- Add admin endpoints + table
- Add endpoint to modify a users tier as an admin
- Add script to create an admin.

An admin can use the CI tool to update a user's role.


web3_proxy/src/frontend/users.rs
```diff
@@ -853,3 +855,135 @@ pub async fn user_stats_detailed_get(
...
+        .context("No user with this id found!")?;
+    // TODO: Let's connect the string, and find the previous string of the user id ... (this might be ok for now too thought)
+    response_body.insert(
+        "previous_tier",
+        serde_json::Value::Number(user.user_tier_id.into()),
+    );
+
+    // Modify the user-role ...
+    // Check if this use has admin privileges ...
+    let user_role: secondary_user::Model = secondary_user::Entity::find()
```
WyseNynja, line 936: `secondary_user` is going to be used for something else. Users will be able to add other users who can view their stats and administrate their account.

Site-wide admins will be on a table that still needs to be created.

This should be a query on the `user` table to find the user that the admin is trying to modify.

web3_proxy/src/frontend/users.rs
```diff
@@ -853,3 +855,135 @@ pub async fn user_stats_detailed_get(
...
+        .context("admin_change_user_roles needs a db replica")?;
+    let mut redis_conn = app
+        .redis_conn()
+        .await
+        .context("admin_change_user_roles had a redis connection error")?
+        .context("admin_change_user_roles needs a redis")?;
+
+    // TODO: Make a single query, where you retrieve the user, and directly from it the secondary user (otherwise we do two jumpy, which is unnecessary)
+    // get the user id first. if it is 0, we should use a cache on the app
+    let user_id = get_user_id_from_params(&mut redis_conn, &db_conn, &db_replica, bearer, &params).await?;
```
WyseNynja, line 915: let's call this `admin_id` and then add a query here that checks that the `admin_id` is in an `admin` table

web3_proxy/src/frontend/users.rs
```diff
@@ -853,3 +855,135 @@ pub async fn user_stats_detailed_get(
...
+
+    let mut response_body = HashMap::new();
+    response_body.insert(
+        "user_id",
+        serde_json::Value::Number(user_id.into()),
+    );
+
+    // Get both the user and user role
+    let user: user::Model = user::Entity::find_by_id(user_id)
+        .one(&db_conn)
```
WyseNynja, line 925: this should use `db_replica`

web3_proxy/src/frontend/mod.rs
```diff
@@ -85,6 +85,7 @@ pub async fn serve(port: u16, proxy_app: Arc<Web3ProxyApp>) -> anyhow::Result<()
             get(users::user_stats_aggregated_get),
         )
         .route("/user/stats/detailed", get(users::user_stats_detailed_get))
+        .route("/user/modify_role", get(users::admin_change_user_roles))
```
WyseNynja, line 88: let's put this at `/admin/modify_role`. Regular users should not have access to do this.

WyseNynja (changes requested):

web3_proxy/src/frontend/users.rs
yenicelik: not entirely sure, but what is the (practical) difference to why we have both a db_replica, and a db_conn? speed?

web3_proxy/src/frontend/users.rs
yenicelik: I called it caller_id (which is later resolved to an admin, if he is truthfully an admin). can change it to admin_id if you prefer

web3_proxy/src/frontend/users.rs
yenicelik: added an admin table (rn without "description", i can add it now, or also at a later migration). as you prefer.

web3_proxy/src/frontend/users.rs
WyseNynja: caller_id is good

web3_proxy/src/frontend/users.rs
WyseNynja: db_replica is read only and replicated all over the world. It's much faster to query since everything doesn't have to go all the way to Ohio. 

If you are just reading the data, use replica.

If you are reading the data to modify and then write it or are writing brand new data, use the main db.

yenicelik:
- Added removing user's login cache (entry in Login). 
- Added script to upgrade a user to an admin through CLI
- Create Admin table.

@WyseNynja again, would be great to have some guidance reg. tests (where and how these should be added).

WyseNynja:
reminder: add a test for this new endpoint

yenicelik:
Will close this PR as the changes are included in https://github.com/llamanodes/web3-proxy/pull/20

Issue #19, opened 2023-01-29
Admin endpoint for imitating a user (read-only)


yenicelik:
Let admins login as another account so that they can diagnose issues with the dashboards etc. This should be a read-only endpoint, i.e. if the admin is logged in as the user, there should be no permission to write data. 

- [ ] Login admin as user
- [ ] Logic to distinguish admin from user whenever a write operation is happening (should be little code if possible)

yenicelik:
Also merged and successfully tested in prod as I understand

PR #20, opened 2023-01-30
Admin endpoint (Godmode / Imitate User)

yenicelik:
#19 
- Modify PendingLogin and Login table
- Add endpoints to login and logout as an admin


migration/src/m20230130_124740_read_only_login_logic.rs
```diff
@@ -0,0 +1,43 @@
...
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Alias::new("login"))
```
WyseNynja, line 13: Nitpick: this should be Admin::Table. Doesn't make a huge difference but I try to only use Alias when I don't have an Iden enum available

migration/src/m20230130_124740_read_only_login_logic.rs
```diff
@@ -0,0 +1,43 @@
...
+                    .add_column(
+                        ColumnDef::new(Login::ReadOnly)
+                            .boolean()
+                            .not_null()
+                    ).to_owned()
+            ).await
+    }
+
+    async fn down(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
```
WyseNynja, line 23: Nitpick: this comment should be removed

migration/src/m20230130_124740_read_only_login_logic.rs
```diff
@@ -0,0 +1,43 @@
...
+        // Drop the column from the table ...
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Alias::new("login"))
+                    .drop_column(Alias::new("read_only"))
```
WyseNynja, line 29: above, Use the Iden enum instead of an Alias

web3_proxy/src/admin_queries.rs
```diff
@@ -0,0 +1,125 @@
...
+    // TODO: Make a single query, where you retrieve the user, and directly from it the secondary user (otherwise we do two jumpy, which is unnecessary)
+    // get the user id first. if it is 0, we should use a cache on the app
+    let caller_id = get_user_id_from_params(&mut redis_conn, &db_conn, &db_replica, bearer, &params).await?;
+
+    // Check if the caller is an admin (i.e. if he is in an admin table)
+    let admin: admin::Model = admin::Entity::find()
+        .filter(admin::Column::UserId.eq(caller_id))
+        .one(&db_conn)
+        .await?
+        .context("This user is not registered as an admin")?;
```
WyseNynja, line 82: I think there's an AccessDenied frontend error that we should use. Using "context" is quick and easy, but it ends up being served as a 500. And 500s get automatically retried multiple times across the autoscaling group.

web3_proxy/src/bin/web3_proxy_cli/main.rs
```diff
@@ -70,6 +71,7 @@ pub struct Web3ProxyCli {
 enum SubCommand {
     ChangeUserAddress(change_user_address::ChangeUserAddressSubCommand),
     ChangeUserTier(change_user_tier::ChangeUserTierSubCommand),
+    ChangeUserAdminStatus(change_user_admin_status::ChangeUserAdminStatusSubCommand),
```
WyseNynja, line 74: Nitpick: alphabetical order please

web3_proxy/src/admin_queries.rs
yenicelik: Ok I think I added it in all the right places again. just lmk if you see a spot where it's off (i.e. supposed to be a 400 instead of a 500 or so)

yenicelik:
Also includes changes for #16

scripts/install-test-suite.sh
```diff
@@ -0,0 +1,3 @@
+cargo install cargo-binstall
+cargo install cargo-binstall
```
WyseNynja, line 2: is this supposed to be here twice?

scripts/apply-migrations.sh
```diff
@@ -0,0 +1,3 @@
+sea-orm-cli migrate up
+
+#  sea-orm-cli generate entity -t <table_name>
```
WyseNynja, line 3: there's a few more flags you need. you also have to change a few column types (i8 -> bool, Vec<u8> -> Uuid, and some others) manually. check the README.

web3_proxy/src/admin_queries.rs
```diff
@@ -0,0 +1,134 @@
...
+    app: &'a Web3ProxyApp,
+    bearer: Option<TypedHeader<Authorization<Bearer>>>,
+    params: &'a HashMap<String, String>
+) -> Result<Response, FrontendErrorResponse> {
+
+    // Quickly return if any of the input tokens are bad
+    let user_address: Vec<u8> = params
+        .get("user_address")
+        .ok_or_else(||
+            FrontendErrorResponse::StatusCode(
```
WyseNynja, line 36: This works, but I made a `FrontendErrorResponse::BadRequest` to make this less verbose.

web3_proxy/src/admin_queries.rs
```diff
@@ -0,0 +1,134 @@
...
+            None,
+        ))?;
+    // Return early if the target user_tier_id is the same as the original user_tier_id
+    response_body.insert(
+        "user_tier_title",
+        serde_json::Value::Number(user.user_tier_id.into()),
+    );
+
+    // Now we can modify the user's tier
+    let new_user_tier: user_tier::Model = !user_tier::Entity::find()
```
WyseNynja, line 104: is there supposed to be a `!` on this? That seems wrong

web3_proxy/src/admin_queries.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        user.save(&db_conn).await?;
+
+        info!("user's tier changed");
+    }
+
+    // Finally, remove the user from redis
+    // TODO: Also remove the user from the redis
+    // redis_conn.zrem();
+    // redis_conn.get::<_, u64>(&user.) // TODO: Where do i find the bearer token ...
```
WyseNynja, line 129: I think you'll need to do a query on the [login](https://github.com/llamanodes/web3-proxy/blob/main/entities/src/login.rs) table. That should let you get all bearer tokens for the user.

I think you'll want to delete them from redis and from the database.

web3_proxy/src/admin_queries.rs
```diff
@@ -0,0 +1,134 @@
...
+            "Unable to get the user_tier_title key from the request".to_string(),
+            None,
+        ))?;
+
+    // Prepare output body
+    let mut response_body = HashMap::new();
+
+    // Establish connections
+    let db_conn = app.db_conn().context("query_admin_modify_user needs a db")?;
+    let db_replica = app
```
WyseNynja, line 64: Since this endpoint writes to the database, do not get `db_replica`. Just use db_conn for all queries

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
```diff
@@ -0,0 +1,73 @@
...
+        let user = user::Entity::find()
+            .filter(user::Column::Address.eq(address.clone()))
+            .one(db_conn)
+            .await?
+            .context("No user found with that address")?;
+
+        // Check if there is a record in the database
+        let mut admin = admin::Entity::find()
+            .filter(admin::Column::UserId.eq(address))
+            .all(db_conn)
```
WyseNynja, line 42: i think this can be `.one`

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
```diff
@@ -0,0 +1,73 @@
...
+            .await?;
+
+        debug!("user: {:#?}", user);
+
+        match admin.pop() {
```
WyseNynja, line 47: if you use `.one` above, you can remove this `pop`

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
```diff
@@ -0,0 +1,73 @@
...
+                    user_id: sea_orm::Set(user.id),
+                    ..Default::default()
+                };
+                new_admin.insert(db_conn).await?;
+            },
+            Some(old_admin) if !should_be_admin => {
+                // User is already an admin, but shouldn't be
+                old_admin.delete(db_conn).await?;
+            },
+            _ => {}
```
WyseNynja, line 60: add a log saying something like `no change needed for {address}`

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        // Remove any user logins from the database (incl. bearer tokens)
+        let delete_result = login::Entity::delete_many()
+            .filter(login::Column::UserId.eq(user.id))
+            .exec(db_conn)
+            .await;
+
+        debug!("cleared modified logins: {:?}", delete_result);
```
WyseNynja, line 69: also clear redis.

since we do need this same code here, in a frontend endpoint, and similar logic in the /logout endpoint (that just clears the one bearer token though), should we make a helper function?

web3_proxy/src/bin/web3_proxy_cli/main.rs
```diff
@@ -293,6 +295,15 @@ fn main() -> anyhow::Result<()> {
 
                 x.main(&db_conn).await
             }
+            SubCommand::ChangeUserAdminStatus(x) => {
+                let db_url = cli_config
+                    .db_url
+                    .expect("'--config' (with a db) or '--db-url' is required to run proxyd");
```
WyseNynja, line 301: change to `required to run change_user_admin_status`

web3_proxy/src/frontend/mod.rs
```diff
@@ -160,6 +161,14 @@ pub async fn serve(port: u16, proxy_app: Arc<Web3ProxyApp>) -> anyhow::Result<()
         )
         .route("/user/stats/detailed", get(users::user_stats_detailed_get))
         .route("/user/logout", post(users::user_logout_post))
+        .route("/admin/modify_role", get(admin::admin_change_user_roles))
+        .route("/admin/imitate-login/:user_address", get(admin::admin_login_get))
+        .route(
+            "/user/imitate-login/:user_address/:message_eip",
```
WyseNynja, line 167: i think this route should be `/admin/imitate-login/:user_address/:message_eip` to match the one before it

WyseNynja (changes requested):

web3_proxy/src/admin_queries.rs
yenicelik: I replaced the majority with read-replicas now. The reason is that there is a case where the user-tier needs not to be modified, in this case writing is unnecessary. I can still go ahead and just use db_conn though if you prefer

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
WyseNynja: i don't think you need to do anything with the salt. the salted stuff is just to keep track that the user has logged in. we use that for our own tracking. we don't want to delete anything from there

WyseNynja: i thought we had the bearer tokens cached in redis. that's what i was referring to. i might have deleted that when we moved everything to the db, but I don't think so.

WyseNynja:
lets add an "audit trail" table that saves all admin logins. we can build alerting or a slack message or something around this.

scripts/manual-tests/16-change-user-tier.sh
```diff
@@ -0,0 +1,38 @@
...
+# Run the proxyd instance
+RUSTFLAGS="--cfg tokio_unstable" cargo run --release -- proxyd
+
+# Check if the instance is running
+curl -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"web3_clientVersion","id":1}' 127.0.0.1:8544
+
+# Login in the user first (add a random bearer token into the database)
+# (This segment was not yet tested, but should next time you run the query)
+INSERT INTO login (bearer_token, user_id, expires_at, read_only) VALUES (
+  "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIiwiaWF0IjoxNTE2MjM5MDIyfQ.SflKxwRJSMeKKF2QT4fwpMeJf36POk6yJV_adQssw5c",
```
WyseNynja, line 21: as discussed in telegram, this line needs to be a valid UUID/ULID

web3_proxy/src/admin_queries.rs
WyseNynja: I think that's safest. I've had races in the past cause problems

scripts/manual-tests/16-change-user-tier.sh
yenicelik: Using SIWE through browser did the trick, also UUID seems like I would have to insert the bytecode for the UUID, string like this most likely won't work, we can fix this when doing CICD in the future

web3_proxy/src/bin/web3_proxy_cli/change_user_admin_status.rs
yenicelik: yh looks like it was deleted in that case. I will remove this segment too then.

web3_proxy/src/admin_queries.rs
yenicelik: `get_user_id_from_params` seems to still be using it, i will not delete it, but replace usage in all of the file (except for this function) with db_conn

yenicelik:
> lets add an "audit trail" table that saves all admin logins. we can build alerting or a slack message or something around this.

added this, requested review

WyseNynja:
I should have some time to review this tomorrow

yenicelik:
@WyseNynja with this and the other PRs probably best to first merge stats_v2 into master. I can also open PRs into stats_v2

WyseNynja:
AFAICT, this pr doesn't touch any of the code that's changing for the stats refactor. I don't think you should block deploying this on me finishing stats v2. Open the PR against main.

Issue #22, opened 2023-02-01
Use tokio tracing instead of log


WyseNynja:
I had tracing in the repo originally, but I was having some performance issues with it using like 50% of the app's time. Rather than fix all the fields and put "instrument" a bunch of places, I just switched to the regular logging crate and string formatting

however, now that things get sent to sentry, this is making tons of entries that are all for the same thing.

Let's put tracing back in and use fields/skip/properly.

yenicelik:
which parts should be skipped? logging everything as trace for now

yenicelik:
Also tracer requires any argument to have the `Debug` trait implemented. Are we ok with wrapping third-party structs and types, or shall we ignore these cases for now?

WyseNynja:
I'm not totally sure which should be skipped. Probably at least ones that don't actually do any logging inside the function. 

For things that don't have debug impls, it's probably simplest enough to just skip them for now. Or prefix the field with % to use the toString

yenicelik:
this can be closed now?

PR #23, opened 2023-02-03
Replace logging by tracing

yenicelik:
#22 
Oriented around: 
- 0e1cf5767c04eaa4cc9322a6c69506cff385ed28
- a534eae9684ccdac57c48ac3067b917114644dc2

- Added a bunch of `Debug` derives
- Check skip_all logic, especially for traits with strong generics
- will run rustfmt if files generally look ok (to order the imports mostly)

### DIFF
--- a/web3_proxy/src/app/mod.rs
+++ b/web3_proxy/src/app/mod.rs
@@ -124,7 +124,7 @@ impl PartialEq for ResponseCacheKey {
 impl Eq for ResponseCacheKey {}
 
 impl Hash for ResponseCacheKey {
-    #[instrument(level = "trace")]
+    #[instrument(skip_all)]
     fn hash<H: Hasher>(&self, state: &mut H) {
         self.block.as_ref().map(|x| x.hash()).hash(state);
         self.method.hash(state);
@@ -1252,7 +1252,7 @@ impl Web3ProxyApp {
                         &request,
                         Some(request_metadata.clone()),
                         None,
-                        Level::Trace,
+                        Level::TRACE,
                         num,
                     )
                     .await?;
--- a/web3_proxy/src/app/mod.rs
+++ b/web3_proxy/src/app/mod.rs
@@ -31,7 +31,7 @@ use futures::future::join_all;
 use futures::stream::{FuturesUnordered, StreamExt};
 use hashbrown::{HashMap, HashSet};
 use ipnet::IpNet;
-use log::{debug, error, info, trace, warn, Level};
+use tracing::{debug, error, info, trace, warn, Level};
 use metered::{metered, ErrorCount, HitCount, ResponseTime, Throughput};
 use migration::sea_orm::{
     self, ConnectionTrait, Database, DatabaseConnection, EntityTrait, PaginatorTrait,
--- a/web3_proxy/src/app/mod.rs
+++ b/web3_proxy/src/app/mod.rs
@@ -31,7 +31,7 @@ use futures::future::join_all;
 use futures::stream::{FuturesUnordered, StreamExt};
 use hashbrown::{HashMap, HashSet};
 use ipnet::IpNet;
-use tracing::{debug, error, info, trace, warn, Level};
+use tracing::{debug, error, info, instrument, trace, warn, Level};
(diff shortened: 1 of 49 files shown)


yenicelik:
@WyseNynja feedback on this would be appreciated https://github.com/llamanodes/web3-proxy/pull/23/files#diff-90681dafe8ccb7c7e7fd88b7e4258790fb9a93a9d17361fa5577263e7a9bc5d7R192 i.e. is this ok as an entrypoint?

yenicelik:
TODO: @yenicelik 
- merge from upstream
- Modify tests accordingly

web3_proxy/src/app/mod.rs
```diff
@@ -79,6 +79,7 @@ struct ResponseCacheKey {
 }
 
 impl ResponseCacheKey {
+    #[instrument(level = "trace")]
```
WyseNynja, line 82: i think putting trace instruments at every function is part of what was making things so slow. so we should definitely benchmark (with ethspam) a build with and a build without and see the impact

web3_proxy/src/app/mod.rs
```diff
@@ -167,11 +170,12 @@ pub struct AuthorizationChecks {
 
 /// Simple wrapper so that we can keep track of read only connections.
 /// This does no blocking of writing in the compiler!
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct DatabaseReplica(pub DatabaseConnection);
 
 // TODO: I feel like we could do something smart with DeRef or AsRef or Borrow, but that wasn't working for me
 impl DatabaseReplica {
+    #[instrument(skip_all)]
```
WyseNynja, line 178: is it worth adding instrument to functions that don't log anything? is that necessary to make sure spans carry over properly? I'm honestly not sure.

web3_proxy/src/bin/wait_for_sync.rs
```diff
@@ -43,7 +45,20 @@ async fn main() -> anyhow::Result<()> {
         std::env::set_var("RUST_LOG", "wait_for_sync=debug");
     }
 
-    env_logger::init();
+    // TODO: how do we put the EnvFilter on this?
```
WyseNynja, line 48: i think this todo is done by `EnvFilter::from_default_env()`

web3_proxy/src/bin/web3_proxy_cli/main.rs
```diff
@@ -187,35 +188,40 @@ fn main() -> anyhow::Result<()> {
...
-
-        Some(guard)
-    } else {
-        log::set_boxed_logger(Box::new(logger)).unwrap();
-
-        None
-    };
-
-    log::set_max_level(max_level);
+    // Setup the tracing subscriber ... (I think this is everything that's needed (?))
```
WyseNynja, line 191: i think we need to do more so that sentry also gets attached

web3_proxy/src/bin/web3_proxy_cli/sentryd/mod.rs
```diff
@@ -125,9 +125,9 @@ impl SentrydSubCommand {
                 }
 
                 while let Some(err) = error_receiver.recv().await {
-                    log::log!(err.level, "check failed: {:#?}", err);
+                    error!("check failed: {:#?}", err);
```
WyseNynja, line 128: `error` is too verbose. Sometimes these are errors and other times they are warns. So this needs to keep using `err.level` (which probably needs a modification)

WyseNynja (changes requested):
This is a good start.

The next big piece is changing all the logs that look like this:

```
debug!("Failed reconnect to {}! Retry in {}ms. err={:?}", self, retry_in.as_millis(), err);
```

To looking something like this (this isn't right, but its the general idea):

```
#[instrument( level = debug, fields = self) ]

debug!(retry_in=retry_in.as_millis(), ?err, "Failed reconnect {}", err);
```

I /think/ we want the error in there twice. once as a field with the debug format so that we get all the context/backtrace. And also in the message (with the basic string formatter) so that sentry doesn't combine different error types into the same entry.

Issue #24, opened 2023-02-03
Referral Program: Generate Referral Code tied to user


yenicelik:
Tables:
- Create a `Referral Code` Table, that is tied to a `User` user_id. Table has following specs:
    - user_ids are unique
- Create a `Referrals` Table, that records any new user who joined, using the respective referral code:
    - (`referree`, `referrer`, `valid_from_state`): both are foreign keys to the User table
- `Request` Table. This table creates a log of every transaction that a user is generating. It does not store requests older than 1 minute
- Create a `Credits` Table. This is only available to premium users, who may have free extra credits (can top up and all that).
    - user_ids are unique
    - credit: BigNumber
- `Transaction` Table. This table is responsible for recording transactions that take up credits. Credits are removed from the user when the respective transaction has enough confirmations (or if the amount of transactions that can be executed is lower than the credit available). 
    - user_id
    - transaction_hash: unique
    - chain: Enum (could also be a string for now)  

New Endpoints:
- /user/referral_code
    - Generates & returns a referral code that the user can now distribute
- /signup/:referral_code
    - Creates an entry in the table that the user is now "registered". The credit counter is initialized
- /user/add-credit
    - Tops up the credit that a user has. Accepts stablecoins, and converts them into credits
- / tx
    - Registers a transaction that was confirmed (basically to remove credits)

Affected Endpoints: 
- /proxy_web3_rpc (all RPC endpoints)
    - also submits a tx_hash s.t. this can (by a subprocess) reduce credit for the submitting user
- /user/balance

Other notes:
- We might have to run a cron job to check if a user has deposited enough credits into their accounts
- We could also have an RPC counter just to double check (before each RPC call), if the referred user's credits have been replenished. Here we can also mutate the user's table after each RPC call, updating it with how many credits were consumed.

Other questions to @WyseNynja : how are credits/funds tracked? And in which currency (if there is any)?

WyseNynja:
We don't have any code for tracking credits/funds yet.

And it's not "transactions" that use credits. It's all rpc requests that use credits. I don't think we want cron job. We should instead check the cached user data to know if they have a positive balance or not. It's cached for 10 minutes. It's fine if someone's balance goes slightly negative from 10 minutes of excess traffic. 

[...]

PR #33, opened 2023-02-16
PR WIP - 24 referral program

yenicelik:
#24 So far worked on 

- User can deposit tokens using the `POST /user/balance/:tx_hash` endpoint (prior to that sending coins to the deposit contract) [not yet tested]
- User can check their balance using the ``GET /user/balance` endpoint
- Added deposit_receipt table
- Added balance table
- Added referrer Table (every user, assuming they're premium can generate a referral code)
- Added referee Table (records what referral code a user has redeemed)
- Added endpoint to create a referral link (assuming premium user)
- WIP endpoint to redeem a referral link. This does not yet mean that the user is provided with credit, but rather that the spend-tracking starts. Also we should prob assume that the referred user should not have been signed up already?

What should prob be removed again:
- rpc_request table

What's missing:
- Spend table (this will also control if the referrer will get their extra free tokens, according to whether the referee has )
- After each spending call, we need to check (and deposit) items to the referrer. Add a column to the referee (recognized) to check whether the referee and referrer have both received credits

Please see this PR and not the issue as there were significant changes to the issue.


yenicelik:
@WyseNynja pointing out any bigger (design) flaws would be helpful

yenicelik:
The workflow looks like this:
(1) X is already a premium user
(2) X creates a referral code
(3) X sends the referral code to Y
(4) Y creates a premium account
(5) Y deposits tokens to receive a balance
(6) Y starts using RPC requests. Each RPC request adds to a "spent" table
(7) After each RPC request, the "spent" table is checked. 
(8) If the spent credits, according to the spent table is above a certain threshold, assign N credits to X, and M credits to Y.
(9) Mark that Y has used up his referral code, they cannot be deposited "free" credits again

WyseNynja:
So we said earlier that a user should only have one referral code. Should we also have it do that once a referral is set, it can't be changed? I think makes the code a lot easier. But maybe not

WyseNynja:
> (8) If the spent credits, according to the spent table is above a certain threshold, assign N credits to X, and M credits to Y.

Why are giving credits to both X and Y?

yenicelik:
> So we said earlier that a user should only have one referral code. Should we also have it do that once a referral is set, it can't be changed? I think makes the code a lot easier. But maybe not

Yes this is the case right now

yenicelik:
Closing as outdated

PR #35, opened 2023-02-26
24 referral program

yenicelik:


web3_proxy/src/accepted_currencies.rs
```diff
@@ -0,0 +1,31 @@
...
+    name: String,
+    address: Address,
+    decimals: u8
+}
+
+const CURRENCIES: HashMap<Address, Currency> = lazy_static! {
+    let mut currencies: HashMap<Address, Currency> = HashMap();
+
+    currencies.insert(
+        "0xdAC17F958D2ee523a2206206994597C13D831ec7".parse::<Address>().unwrap(),
```
yenicelik, line 13: @WyseNynja I suppose we will only have a single currency that we accept?

web3_proxy/src/balance.rs
```diff
@@ -0,0 +1,12 @@
+use migration::sea_orm::DatabaseConnection;
+use crate::frontend::errors::FrontendErrorResponse;
+
+/// Implements logic to include balances
+///
+/// Adds an entry to the `spent` table (rpc_requests)
+/// Then subtracts the user balance from the `balance` table
+pub async fn update_balance_and_spend(user_id: u64, db_conn: &DatabaseConnection, amount_delta: u64) -> Result<u64, FrontendErrorResponse> {
```
yenicelik, line 8: added this as a todo, in case there is more complex / modular logic to be implemented. don't know exactly where to place this yet though (I would assume into the /rpc/ endpoints)

web3_proxy/src/referral_code.rs
```diff
@@ -0,0 +1,25 @@
+use anyhow::{self, Result};
+use thread_fast_rng::rand::distributions::Alphanumeric;
+use thread_fast_rng::rand::{Rng, thread_rng};
+
+pub struct ReferralCode(pub String);
+
+impl Default for ReferralCode {
+    fn default() -> Self {
+        let mut rng = thread_rng();
+        let chars: String = (0..32).map(|_| rng.sample(Alphanumeric) as char).collect();
```
yenicelik, line 10: it's not exactly readable, but im assuming here that codes are shared via link (which they usually are to my knowledge..) can look into more readable formats if you prefer

yenicelik:
@WyseNynja stats-v2 and main seem somewhat divergent, and merging from both leads to this branch including a bit of both. perhaps good to consolidate on branches soon?

web3_proxy/src/referral_code.rs
WyseNynja: Use ULIDs. The crate is already used by web3-proxy. They are short and readable.

web3_proxy/src/balance.rs
WyseNynja: This should probably be done in with the code that aggregates and saves data to the rpc_accounting_v2 table. I don't think we need a separate "spent" table. It should be fine to just add a column to rpc_accounting. And it's vital that we aggregate these spends. We do billions of requests and if they aren't aggregated, the db is going to become seriously bloated.

web3_proxy/src/accepted_currencies.rs
WyseNynja: No. Launch will include multiple stablecoins. 

I am pretty sure the factory contract has a list of accepted currencies already. We should query that rather than hard coding here. We don't want to be forced to sync deploys with changes to the smart contracts. 

Also, tokens bridged to other chains have a different address

WyseNynja:
> @WyseNynja stats-v2 and main seem somewhat divergent, and merging from both leads to this branch including a bit of both. perhaps good to consolidate on branches soon?

Why are you merging from both? If you just merge from stats_v2, I think this will be much simpler. I'll merge main into stats v2 before deploying it.

web3_proxy/src/referral_code.rs
yenicelik: https://github.com/llamanodes/web3-proxy/pull/35/commits/edd9371284554de7e5ed083cd0d03169a76df791

web3_proxy/src/accepted_currencies.rs
yenicelik: I think we will still need some hard-coded list to convert from absolute token amounts to decimals (i.e. stablecoins can have different decimal points, might be needed for conversion)

yenicelik: would probably need something like this: https://github.com/Uniswap/v2-core/blob/master/contracts/UniswapV2Pair.sol#L38 in the payments contract to not have to read them from the storage slots directly

WyseNynja:
You can query that at start, too. It's a simple eth_call with the standard ERC20 abiOn Mar 3, 2023, at 5:04 AM, David ***@***.***> wrote:﻿
@yenicelik commented on this pull request.



In web3_proxy/src/accepted_currencies.rs:
> @@ -0,0 +1,31 @@
+/// Accepted Currencies, their respective ERC20 token function, data and decimals
+
+pub struct Currency {
+    name: String,
+    address: Address,
+    decimals: u8
+}
+
+const CURRENCIES: HashMap<Address, Currency> = lazy_static! {
+    let mut currencies: HashMap<Address, Currency> = HashMap();
+
+    currencies.insert(
+        "0xdAC17F958D2ee523a2206206994597C13D831ec7".parse::<Address>().unwrap(),

I think we will still need some hard-coded list to convert from absolute token amounts to decimals (i.e. stablecoins can have different decimal points, might be needed for conversion)

web3_proxy/src/balance.rs
yenicelik: I added them here, does this look about right? https://github.com/llamanodes/web3-proxy/pull/35/commits/ea3125e7d4c7392dcc26f8d8d2e76a9ef3bde3ef

yenicelik:
closing at outdated

PR #44, opened 2023-04-12
User Balance + Referral Logic

yenicelik:
User balance addition:
- added deposit_contract and deposit_topic to the config/example.toml
added config/development_polygon.toml (to save some work adding RPC providers, feel free to delete this in a future revision / at this point, i can post it here as a comment)
- Decimal digits before and after the comma are specified (20 before, 10 after for now)
- added brownie test suite to simplify reproduction of depositing to payment contracts. Contracts are not needed as the data is read from polygonscan
- put payment deposit logic into frontend/payment.rs . withdrawal will also be more complex (most likely), so a separate file is probably justified
- added logic to compute cost from inbound bytes, outbound bytes, and whether or not the response result is in the cache
- balance table tracks the historically used_balance, and the currently available_balance
- increase_balance_receipt
- renamed some variables in users.rs to make it more readable
- added credits used to stats fields
- added arithmetic to calculate cost by bytes
- endpoint to see all deposited transactions
- stats endpoint allows for historical balance tracking if aggregated / detailed is opted in

Referral logic:
- added tables "referee" and "referrer", capturing referral code, who referred who, if the 50$ were applied, and how much credits the referrer received from the referee's usage
- added sum_credits_used to rpc_accounting_v2 for statistics purposes
- 10% are provided to the referrer (from the referee's spend), for 1 year after the invite link was updated. time of start is recorded in the database
- added (manual) bash script to test referral logic
- Added error message "InvalidReferralCode"
- "/user/referral", get(users::user_referral_link_get) fetches a user's referral code
- generate a referral code that looks like "llamanodes-ULID". ULID may expose date-information, but that should be fine for privacy concerns (and perhaps even intended at a later stage / anti-spam detection)
- if the user spends 100$, they get 100$ on top

Secondary user logic:
- RPC key that can be shared (this marks a project). This was added in entities
- Test case to check that (1) primary user can share RPC keys with secondary users (2) primary user can see RPC keys were shared (3) secondary user can see all the RPC keys he has access to

Downgrade premium user logic:
- Added a "premium" and "premium out of funds" tier
- Added column "downgrade to", which is downgraded to when funds run out (right now only for premium)

refactored some of the user.rs file into users/....rs (due to file starting to become very large)


web3_proxy/src/lib.rs
```diff
@@ -30,4 +31,5 @@ pub struct PostLoginQuery {
 pub struct PostLogin {
     sig: String,
     msg: String,
+    pub referral_code: Option<String>,
```
yenicelik, line 34: @WyseNynja was not sure how to retrieve the referral_code from the other struct, do you think leaving it here is fine? happy to also move it

yenicelik:
TODO david:
- Make unique constraint in IncreaseBalanceReceipt based on (ChainId, TxHash), not only (TxHash)
- TODO: Remove daemon.rs changes (I didn't touch them, the ones on the devel branch should persist. Could be that these come from "main", which was merged a while ago.

yenicelik:
related to https://github.com/llamanodes/web3-proxy/issues/24

yenicelik:
TODO: Add premium tier (id=5)

web3_proxy/src/app/mod.rs
```diff
@@ -608,6 +608,104 @@ impl Web3ProxyApp {
                 .build()?,
         );
 
+        // // TODO: i don't like doing Block::default here! Change this to "None"?
+        // let (watch_consensus_head_sender, watch_consensus_head_receiver) = watch::channel(None);
```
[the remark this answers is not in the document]
WyseNynja: yeah delete both. lines

web3_proxy/src/bin/web3_proxy_cli/daemon.rs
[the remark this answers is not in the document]
WyseNynja: Hmm. I'm not sure. Maybe just copy/paste whats on main right now and see if any changes show up

web3_proxy/src/stats/mod.rs
```diff
@@ -435,6 +664,7 @@ impl StatBuffer {
         // any errors inside this task will cause the application to exit
         let handle = tokio::spawn(async move {
             new.aggregate_and_save_loop(bucket, stat_receiver, shutdown_receiver)
+                // new.aggregate_and_save_loop(stat_receiver, shutdown_receiver)
```
WyseNynja, line 692: i think you can delete this commented out code

web3_proxy/src/accepted_currencies.rs
```diff
@@ -0,0 +1,31 @@
...
+        "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48".parse::<Address>().unwrap(),
+        {
+            name: "USDC",
+            address: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48".parse::<Address>().unwrap(),
+            decimals: 6
+        }
+    );
+
+    currencies
+};
```
WyseNynja, line 31: this is fine for right now. But we need to change this to query the chain and get the correct data

web3_proxy/src/rpcs/request.rs
```diff
@@ -27,7 +27,7 @@ pub enum OpenRequestResult {
 
 /// Make RPC requests through this handle and drop it when you are done.
 /// Opening this handle checks rate limits. Developers, try to keep opening a handle and using it as close together as possible
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
WyseNynja, line 30: This cannot be changed to clone. This will not clone properly and will cause problems with rate limiting and active request counting. Why did you add this?

web3_proxy/src/stats/mod.rs
```diff
@@ -377,6 +599,13 @@ impl RpcQueryStats {
         let response_millis = metadata.start_instant.elapsed().as_millis() as u64;
         let response_bytes = response_bytes as u64;
 
+        // TODO: Gotta make the arithmetic here
+
+        // TODO: Depending on the method, metadata and response bytes, pick a different number of credits used
+        // This can be a slightly more complex function as we ll
+        // TODO: Here, let's implement the formula
+        let credits_used = Self::compute_cost(request_bytes, response_bytes, backend_requests == 0);
```
WyseNynja, line 607: this will need the method, too

WyseNynja (changes requested):

web3_proxy/src/accepted_currencies.rs
yenicelik: This variable is not used anymore, everything is fetched from chain

web3_proxy/src/stats/mod.rs
yenicelik: I'll add the argument to the function signature, but for now the arithmetic does not depend on method (method is not used for calculation). Lmk if this should change, easily changeable later.

web3_proxy/src/rpcs/request.rs
yenicelik: I was using the same request handle to make multiple requests (within payment). I changed it now to spawn multiple handles (one per request)

web3_proxy/src/bin/web3_proxy_cli/daemon.rs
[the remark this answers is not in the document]
yenicelik: looks like the file was deleted completely... i saw you deleted it in a commit, i will not act further on this

web3_proxy/src/app/mod.rs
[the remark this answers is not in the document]
yenicelik: @WyseNynja I was referring to these lines here in app/mod.rs:

```
// // TODO: i don't like doing Block::default here! Change this to "None"?
        // let (watch_consensus_head_sender, watch_consensus_head_receiver) = watch::channel(None);
        // // TODO: will one receiver lagging be okay? how big should this be?
        // let (pending_tx_sender, pending_tx_receiver) = broadcast::channel(256);
        //
        // // TODO: use this? it could listen for confirmed transactions and then clear pending_transactions, but the head_block_sender is doing that
        // // TODO: don't drop the pending_tx_receiver. instead, read it to mark transactions as "seen". once seen, we won't re-send them?
        // // TODO: once a transaction is "Confirmed" we remove it from the map. this should prevent major memory leaks.
        // // TODO: we should still have some sort of expiration or maximum size limit for the map
        // drop(pending_tx_receiver);
        //
        // // TODO: capacity from configs
        // // all these are the same size, so no need for a weigher
        // // TODO: ttl on this? or is max_capacity fine?
        // let pending_transactions = Cache::builder()
        //     .max_capacity(10_000)
        //     // TODO: different chains might handle this differently
        //     // TODO: what should we set? 5 minutes is arbitrary. the nodes themselves hold onto transactions for much longer
        //     .time_to_idle(Duration::from_secs(300))
        //     .build_with_hasher(hashbrown::hash_map::DefaultHashBuilder::default());
        //
        // // keep 1GB/5 minutes of blocks in the cache
        // // TODO: limits from config
        // // these blocks don't have full transactions, but they do have rather variable amounts of transaction hashes
        // // TODO: how can we do the weigher better?
        // let block_map: BlockHashesCache = Cache::builder()
        //     .max_capacity(1024 * 1024 * 1024)
        //     .weigher(|_k, v: &Web3ProxyBlock| {
        //         // TODO: is this good enough?
        //         1 + v.block.transactions.len().try_into().unwrap_or(u32::MAX)
        //     })
        //     // TODO: what should we set? 5 minutes is arbitrary. the nodes themselves hold onto transactions for much longer
        //     .time_to_idle(Duration::from_secs(300))
        //     .build_with_hasher(hashbrown::hash_map::DefaultHashBuilder::default());
        //
        // // connect to the load balanced rpcs
        // let (balanced_rpcs, balanced_handle, consensus_connections_watcher) = Web3Rpcs::spawn(
        //     block_map.clone(),
        //     top_config.app.chain_id,
        //     db_conn.clone(),
        //     http_client.clone(),
        //     top_config.app.max_block_age,
        //     top_config.app.max_block_lag,
        //     top_config.app.min_synced_rpcs,
        //     top_config.app.min_sum_soft_limit,
        //     pending_transactions.clone(),
        //     Some(pending_tx_sender.clone()),
        //     vredis_pool.clone(),
        //     balanced_rpcs,
        //     Some(watch_consensus_head_sender),
        // )
        // .await
        // .context("spawning balanced rpcs")?;
        //
        // // save the handle to catch any errors
        // cancellable_handles.push(balanced_handle);
        //
        // // connect to the private rpcs
        // // only some chains have this, so this is optional
        // let private_rpcs = if private_rpcs.is_empty() {
        //     // TODO: do None instead of clone?
        //     warn!("No private relays configured. Any transactions will be broadcast to the public mempool!");
        //     None
        // } else {
        //     let (private_rpcs, private_handle, _) = Web3Rpcs::spawn(
        //         block_map,
        //         top_config.app.chain_id,
        //         db_conn.clone(),
        //         http_client.clone(),
        //         // private rpcs don't get subscriptions, so no need for max_block_age or max_block_lag
        //         None,
        //         None,
        //         0,
        //         0,
        //         pending_transactions.clone(),
        //         // TODO: subscribe to pending transactions on the private rpcs? they seem to have low rate limits, but they should have
        //         None,
        //         vredis_pool.clone(),
        //         private_rpcs,
        //         // subscribing to new heads here won't work well. if they are fast, they might be ahead of balanced_rpcs
        //         // they also often have low rate limits
        //         // however, they are well connected to miners/validators. so maybe using them as a safety check would be good
        //         // TODO: but maybe we could include privates in the "backup" tier
        //         None,
        //     )
        //     .await
        //     .context("spawning private_rpcs")?;
        //
        //     if private_rpcs.by_name.is_empty() {
        //         None
        //     } else {
        //         // save the handle to catch any errors
        //         cancellable_handles.push(private_handle);
        //
        //         Some(private_rpcs)
        //     }
        // };
    ```
    
    I am removing these comments for now, it seems repeated below (uncommented)

**Comment by yenicelik** [comment]
I will add the second flux query and fix the merge conflicts tomorrow, some more feedback on recent changes would be appreciated

**Comment by yenicelik** [comment]
@WyseNynja ready to be reviewed and merged

File: config/development_polygon.toml [hunk 7]
```diff
@@ -0,0 +1,216 @@
...
+
+# a database is optional. it is used for user authentication and accounting
+# TODO: how do we find the optimal db_max_connections? too high actually ends up being slower
+db_max_connections = 20
+# development runs cargo commands on the host and so uses "mysql://root:dev_web3_proxy@127.0.0.1:13306/dev_web3_proxy" for db_url
+# production runs inside docker and so uses "mysql://root:web3_proxy@db:3306/web3_proxy" for db_url
+db_url = "mysql://root:dev_web3_proxy@127.0.0.1:13306/dev_web3_proxy"
+
+# deposit_contract = "0xc9fcfa7e28ff320c49967f4522ebc709aa1fde7c"
+deposit_contract = "0x4e3bc2054788de923a04936c6addb99a05b0ea36"
```
WyseNynja, line 12: this has the contract, but no topic. is the topic still needed?

config/development_polygon.toml
```diff
@@ -0,0 +1,216 @@
...
+[app.allowed_origin_requests_per_period]
+"https://chainlist.org" = 1_000
+
+[balanced_rpcs]
+
+[balanced_rpcs.llama_public]
+disabled = false
+display_name = "LlamaNodes"
+http_url = "https://polygon.llamarpc.com"
+#ws_url = "wss://polygon.llamarpc.com/"
```
WyseNynja, line 69: should generally do both http and ws. it subscribes to new heads which is extra important on chains with low block times. otherwise you'll be behind a lot

config/development_polygon.toml
```diff
@@ -0,0 +1,216 @@
...
+[balanced_rpcs.bor]
+disabled = true
+display_name = "bor"
+http_url = "https://polygon-bor.publicnode.com"
+soft_limit = 10
+tier = 2
+
+[balanced_rpcs.blxr]
+disabled = false
+display_name = "blxr"
```
WyseNynja, line 212: doesn't really matter, but having the "display_name" is only useful if its different then the name in the key (both here are blxr)

config/example.toml
```diff
@@ -11,14 +11,18 @@ db_url = "mysql://root:dev_web3_proxy@127.0.0.1:13306/dev_web3_proxy"
 # read-only replica useful when running the proxy in multiple regions
 db_replica_url = "mysql://root:dev_web3_proxy@127.0.0.1:13306/dev_web3_proxy"
 
+# This is the sweeper contract to deposit
```
WyseNynja, line 14: wait. we shouldn't be hard coding a sweeper contract here. we should be putting the factory in the config and then we can get the user's owned copy of the sweeper contract from there

entities/src/increase_balance_receipt.rs
```diff
@@ -0,0 +1,22 @@
...
+use sea_orm::entity::prelude::*;
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize, Deserialize)]
+#[sea_orm(table_name = "increase_balance_receipt")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    #[sea_orm(unique)]
+    pub tx_hash: String,
```
WyseNynja, line 12: should we have this be an Option and then we can use the table for admin and referral credits, too?

entities/src/increase_balance_receipt.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub chain_id: String,
```
WyseNynja, line 13: chain ids should be a u64 (though they are actually smaller than a u64, i think its close enough)

entities/src/increase_balance_receipt.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub amount: Decimal,
+    pub deposit_to_user_id: u64,
```
WyseNynja, line 15: foreign key?

entities/src/increase_balance_receipt.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// TODO: Implement foreign key deposit_to_user_id
```
WyseNynja, line 18: you should do this todo

entities/src/referee.rs
```diff
@@ -0,0 +1,50 @@
+//! `SeaORM` Entity. Generated by sea-orm-codegen 0.10.6
+
+use sea_orm::entity::prelude::*;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "referee")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub used_referral_code: String,
```
WyseNynja, line 10: should this be a foreign key to the referrer table? then we don't need the user id on this table (though it does make the queries simpler)

entities/src/secondary_user.rs
```diff
@@ -11,6 +11,7 @@ pub struct Model {
     pub id: u64,
     pub user_id: u64,
     pub description: Option<String>,
+    pub rpc_key: u64,
```
WyseNynja, line 14: lets keep the name consistent. we use "rpc_key" and "rpc_key_id" and "rpc_secret_key_id". i like "secret" in it because it feels safest

WyseNynja (changes requested):
still need to review more, but i'm stepping away from the computer for a bit

scripts/brownie-tests/contracts/Sweeper.vy
```diff
@@ -0,0 +1,43 @@
+# @version 0.3.7
```
WyseNynja, line 1: Let's not include our contracts here. These are currently in a private repo and I don't think they are ready to release.

web3_proxy/Cargo.toml
```diff
@@ -46,10 +46,12 @@ gethostname = "0.4.2"
 glob = "0.3.1"
 handlebars = "4.3.6"
 hashbrown = { version = "0.13.2", features = ["serde"] }
+hex_fmt = "0.3.0"
```
WyseNynja, line 49: What do we need hexfmt for? The ethers types should handle conversion to hex for us

web3_proxy/src/config.rs
```diff
@@ -94,6 +94,14 @@ pub struct AppConfig {
     /// None = allow all requests
     pub default_user_max_requests_per_period: Option<u64>,
 
+    /// Default ERC address for out deposit contract
+    pub deposit_contract: String,
```
WyseNynja, line 98: These should be optional. Not everyone who runs the proxy will run it with payments. Also, we might be able to use ether's Address type here

web3_proxy/src/config.rs
```diff
@@ -94,6 +94,14 @@ pub struct AppConfig {
...
+
+    /// Default ERC address for out deposit contract
+    pub deposit_topic: String,
```
WyseNynja, line 101: As above, this should be an Option. And I think we can use ether's H256 type on this one

web3_proxy/src/config.rs
```diff
@@ -94,6 +94,14 @@ pub struct AppConfig {
...
+
+    // Accepted ERC20 tokens for the deposit contract
+    // pub accepted_deposit_tokens: Vec<String>,
```
WyseNynja, line 104: We can delete these from config.

web3_proxy/src/frontend/users/referral.rs
```diff
@@ -0,0 +1,93 @@
...
+        .filter(user_tier::Column::Id.eq(user.user_tier_id))
+        .one(db_replica.conn())
+        .await?
+        .ok_or(Web3ProxyError::BadRequest(
+            "Could not find user in db although bearer token is there!".to_string(),
+        ))?;
+
+    warn!("User tier is: {:?}", user_tier);
+    // TODO: This shouldn't be hardcoded. Also, it should be an enum, not sth like this ...
+    if user_tier.id != 6 {
```
WyseNynja, line 53: What if we check user balance instead?

or maybe add a book column to the "out of funds" tier that we can check?

web3_proxy/src/frontend/users/referral.rs
```diff
@@ -0,0 +1,93 @@
...
+        .db_replica()
+        .context("getting replica db for user's revert logs")?;
+
+    // Second, check if the user is a premium user
+    let user_tier = user_tier::Entity::find()
+        .filter(user_tier::Column::Id.eq(user.user_tier_id))
+        .one(db_replica.conn())
+        .await?
+        .ok_or(Web3ProxyError::BadRequest(
+            "Could not find user in db although bearer token is there!".to_string(),
```
WyseNynja, line 48: Since this should be impossible, this might be worth an error.

web3_proxy/src/frontend/users/referral.rs
```diff
@@ -0,0 +1,93 @@
...
+        ))?;
+
+    warn!("User tier is: {:?}", user_tier);
+    // TODO: This shouldn't be hardcoded. Also, it should be an enum, not sth like this ...
+    if user_tier.id != 6 {
+        return Err(
+            anyhow::anyhow!("User is not premium. Must be premium to create referrals.").into(),
```
WyseNynja, line 55: Http 402 payment required

web3_proxy/src/frontend/users/referral.rs
```diff
@@ -0,0 +1,93 @@
...
+        .filter(referrer::Column::UserId.eq(user.id))
+        .one(db_replica.conn())
+        .await?;
+
+    let (referral_code, status_code) = match user_referrer {
+        Some(x) => (x.referral_code, StatusCode::OK),
+        None => {
+            // Connect to the database for mutable write
+            let db_conn = app.db_conn().context("getting db_conn")?;
+
```
WyseNynja, line 70: Should we put a .as_str() on this? Maybe doesn't matter

web3_proxy/src/frontend/users/referral.rs
```diff
@@ -0,0 +1,93 @@
...
+            let referral_code = ReferralCode::default().0;
+            let txn = db_conn.begin().await?;
+            // Log that this guy was referred by another guy
+            // Do not automatically create a new
+            let referrer_entry = referrer::ActiveModel {
+                user_id: sea_orm::ActiveValue::Set(user.id),
+                referral_code: sea_orm::ActiveValue::Set(referral_code.clone()),
+                ..Default::default()
+            };
+            referrer_entry.insert(&txn).await?;
```
WyseNynja, line 80: Is this worth a txn? Looks like there's just the one query inside it

WyseNynja (commented):
Still more to review

config/development_polygon.toml
```diff
@@ -0,0 +1,216 @@
...
+volatile_redis_max_connections = 20
+# development runs cargo commands on the host and so uses "redis://127.0.0.1:16379/" for volatile_redis_url
+# production runs inside docker and so uses "redis://redis:6379/" for volatile_redis_url
+volatile_redis_url = "redis://127.0.0.1:16379/"
+
+# redirect_public_url is optional
+redirect_public_url = "https://llamanodes.com/public-rpc"
+# redirect_rpc_key_url is optional
+# it only does something if db_url is set
+redirect_rpc_key_url = "https://llamanodes.com/dashboard/keys?key={{rpc_key_id}}"
```
WyseNynja, line 38: We need to work with frontend team on this url again. I think they have a different format. Can you make a GitHub issue for this?

entities/src/user_tier.rs
```diff
@@ -11,6 +11,7 @@ pub struct Model {
     pub title: String,
     pub max_requests_per_period: Option<u64>,
     pub max_concurrent_requests: Option<u32>,
+    pub downgrade_tier_id: Option<u64>,
```
WyseNynja, line 14: This should probably have a relation

web3_proxy/src/frontend/users/rpc_keys.rs
```diff
@@ -0,0 +1,259 @@
...
+    let (user, _semaphore) = app.bearer_is_authorized(bearer).await?;
+
+    let db_replica = app
+        .db_replica()
+        .web3_context("getting db for user's keys")?;
+
+    let mut uk = if let Some(existing_key_id) = payload.key_id {
+        // get the key and make sure it belongs to the user
+        rpc_key::Entity::find()
+            .filter(rpc_key::Column::UserId.eq(user.id))
```
WyseNynja, line 104: I think we need something for secondary users with owner/admin permissions here

web3_proxy/src/frontend/users/stats.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    response.insert("page", json!(page));
+    response.insert("page_size", json!(page_size));
+    response.insert("chain_id", json!(chain_id));
+    response.insert("query_start", json!(query_start.timestamp() as u64));
+
+    let db_replica = app
+        .db_replica()
+        .web3_context("getting replica db for user's revert logs")?;
+
```
WyseNynja, line 49: Secondary users should be able to see revert logs, too

web3_proxy/src/rpcs/request.rs
WyseNynja: We should still remove Clone on this though.

web3_proxy/src/stats/mod.rs
```diff
@@ -301,12 +323,212 @@ impl BufferedRpcQueryStats {
...
+            .await?
+            .context(format!(
+                "The foreign key for the user's user_tier_id was not found! {:?}",
+                downgrade_user.user_tier_id
+            ))?;
+
+        // Downgrade a user to premium - out of funds if there's less than 10$ in the account, and if the user was premium before
+        if new_available_balance < Decimal::from(10u64) && downgrade_user_role.title == "Premium" {
+            let mut active_downgrade_user = downgrade_user.into_active_model();
+            active_downgrade_user.user_tier_id = sea_orm::Set(downgrade_user_role.id);
```
WyseNynja, line 421: Wait. Do we actually change the role in the db? I don't think we want that. It's likely we will have multiple roles that all downgrade to the same place. But then we won't know how to restore their roll when they deposit new funds. 

Instead, when we fetch the cached user data, we should set the user tier id there to be the downgraded if their balance is 0

WyseNynja (changes requested):

WyseNynja (changes requested):
I think I've reviewed all of it now

web3_proxy/src/rpcs/request.rs
yenicelik: ah yes fixed the clone dependency but forgot to remove it from the derive, I removed this now

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 members = [
   "deferred-rate-limiter",
   "entities",
+  "influxdb2",
```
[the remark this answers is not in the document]
yenicelik: I removed it from the workspace (included it for debugging back then), there is a github dependency on our fork in web3proxy:
```
influxdb2 = { git = "https://github.com/llamanodes/influxdb2", features = ["rustls"] }
influxdb2-structmap = { git = "https://github.com/llamanodes/influxdb2/"}
```

config/development_polygon.toml
yenicelik: where can i get the ws address from? I was using chainlist to retrieve the http endpoints, i dont see wss there though. (and it's not just always replacing https with wss)

config/development_polygon.toml
yenicelik: ah ok good to know. i kept the names that i capitalized and removed the ones that have case-matching character-sequences

config/example.toml
yenicelik: ah yes this is outdated, i removed it now

entities/src/increase_balance_receipt.rs
yenicelik: i think mixing up the table with referral credits would probably make it too convoluted. i thought about making tx-hash optional and allowing admin credits, but i dont see why we shouldnt just add another table, it will make the data more consistent (and disallow double-spending on the database layer in case any bugs make their way in the future). your call, I can also change it, but i think it'd prefer keeping three separate tables.

scripts/brownie-tests/contracts/Sweeper.vy
yenicelik: oops, my bad! should i try to remove them from the commit, or is it fine to just not merge them into main (technically this one is public now..)

web3_proxy/Cargo.toml
yenicelik: This is where I use it (in frontend/users/payments.rs). I was having troubles using the ethers types HexFmt(keccak256("get_approved_tokens()".to_owned().into_bytes())). I can fiddle around with it again. Let me know what you prefer (and if you happen to know a quick way to format this properly)

web3_proxy/src/frontend/users/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    let mut user: user::ActiveModel = user.into();
+
+    // update the email address
+    if let Some(x) = payload.email {
+        // TODO: only Set if no change
+        if x.is_empty() {
+            user.email = sea_orm::Set(None);
+        } else {
+            // TODO: do some basic validation
```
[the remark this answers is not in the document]
yenicelik: I added this issue, will close this response https://github.com/llamanodes/web3-proxy/issues/60

entities/src/increase_balance_receipt.rs
yenicelik: done

config/development_polygon.toml
yenicelik: good catch, needed indeed, added it

web3_proxy/Cargo.toml
WyseNynja: Leave it for now and make a task. If it works, its fine for now. I made a quick test and this is much shorter and works for me. No need to go from &str -> String -> bytes. keccack256 treats str as utf-8 strings.

Both your code and this code give `0x79aaf078077c817b5d573b12ae04125ffe54cd32f0b7afeede391e27ee973ee9` (thought printing with json wraps in quotes since its a string and thats how JSON prints strings)

```
import ethers::prelude::H256;

let hash = H256::from(keccak256("get_approved_tokens()"));

println!("{:?}", hash);

println!("{}", serde_json::to_string(&hash).unwrap());

```

IIRC, you were trying to avoid using the "debug" formatter. But that is just how the ethers team decided to use it.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -771,6 +778,7 @@ impl Web3ProxyApp {
                             max_requests_per_period: user_tier_model.max_requests_per_period,
                             private_txs: rpc_key_model.private_txs,
                             proxy_mode,
+                            balance: Some(balance),
```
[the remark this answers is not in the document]
WyseNynja: Lets handle this in another task. We aren't going to block release on secondary users

web3_proxy/src/frontend/users/rpc_keys.rs
```diff
@@ -0,0 +1,259 @@
...
+pub async fn rpc_keys_get(
+    Extension(app): Extension<Arc<Web3ProxyApp>>,
+    TypedHeader(Authorization(bearer)): TypedHeader<Authorization<Bearer>>,
+) -> Web3ProxyResponse {
+    let (user, _semaphore) = app.bearer_is_authorized(bearer).await?;
+
+    let db_replica = app
+        .db_replica()
+        .web3_context("db_replica is required to fetch a user's keys")?;
+
```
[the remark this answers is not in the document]
WyseNynja: Secondary users will be handled in another task.

web3_proxy/src/frontend/users/rpc_keys.rs
WyseNynja: Secondary users will be handled in another task.

web3_proxy/src/frontend/users/stats.rs
WyseNynja: Secondary users will be handled in another task.

WyseNynja (changes requested):
we are getting close! Exciting!

PR #68, opened 2023-05-12
use kanal instead of flume or tokio channels

WyseNynja:
it uses unsafe, and its pre-release, but in my testing i get less errors and it seems to be slightly faster


WyseNynja:
oops. kanal isn't cancel safe. need to change the stat sender. the rest are all fine

PR #72, opened 2023-05-14
Added admin balance increase, and also test

yenicelik:
- Admin can add balance to a user
- Created table to track balance additions. 
- Technically negative additions work, but rn automatic downgrade does not work (bcs withdrawal logic is not well-described, could be done in a next iteration)
- user is automatically upgraded to premium if balance is above 10.0
- added short test in bash


WyseNynja:
Why is the automatic downgrade logic blocked on withdrawal logic? A balance going down from requests shouldn't be any different to the code compared to a larger decrease from a withdrawal.

And when the request is made, we load the balance and store it in the authorization checks. The check seeing a negative balance should be enough to change the user to using the downgrade tier. I think that part needs to be finished still, but it's unrelated to withdrawals as far as i can tell

entities/src/admin_increase_balance_receipt.rs
```diff
@@ -0,0 +1,48 @@
...
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "admin_increase_balance_receipt")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    #[sea_orm(column_type = "Decimal(Some((20, 10)))")]
+    pub amount: Decimal,
+    pub admin_id: u64,
+    pub deposit_to_user_id: u64,
```
WyseNynja, line 13: lets add a column for admin notes

web3_proxy/src/frontend/admin.rs
```diff
@@ -33,6 +39,117 @@ use std::sync::Arc;
...
+pub async fn admin_increase_balance(
+    Extension(app): Extension<Arc<Web3ProxyApp>>,
+    TypedHeader(Authorization(bearer)): TypedHeader<Authorization<Bearer>>,
+    Query(params): Query<HashMap<String, String>>,
+) -> Web3ProxyResponse {
+    println!("Got here 1 {:?}", bearer);
+    let (caller, _) = app.bearer_is_authorized(bearer).await?;
+    let caller_id = caller.id;
+
+    println!("Got here 2, {:?}", caller_id);
```
WyseNynja, line 56: clean these prints up

web3_proxy/src/frontend/admin.rs
```diff
@@ -33,6 +39,117 @@ use std::sync::Arc;
...
+        .ok_or_else(|| {
+            Web3ProxyError::BadRequest("Unable to find user_address key in request".to_string())
+        })?
+        .parse::<Address>()
+        .map_err(|_| {
+            Web3ProxyError::BadRequest("Unable to parse user_address as an Address".to_string())
+        })?;
+    let user_address_bytes: Vec<u8> = user_address.clone().to_fixed_bytes().into();
+    // Get the amount from params
+    let amount: Decimal = Decimal::from_str(params.get("amount").ok_or_else(|| {
```
WyseNynja, line 82: this works, but I usually keep it flatter by using `.map(Decimal::from_str)` instead.

web3_proxy/src/frontend/admin.rs
```diff
@@ -33,6 +39,117 @@ use std::sync::Arc;
...
+
+    println!("New amount: {:?}", new_amount);
+
+    // Finally make the user premium if balance is above 10$
+    let premium_user_tier = user_tier::Entity::find()
+        .filter(user_tier::Column::Title.eq("Premium"))
+        .one(&txn)
+        .await?
+        .context("Premium tier was not found!")?;
+    if new_amount >= Decimal::new(10, 0) && user_entry.user_tier_id != premium_user_tier.id {
```
WyseNynja, line 137: I think I commented about this somewhere else. Let's not actually change the user tier in the db. the db should always show the premium tier. if they are out of funds, the rust code should change to using the downgrade_tier_id isntead. 

if you change the database, then once we start adding custom tiers for daos. when they run out of funds, putting them back to the correct tier will be a bigger challenge.

WyseNynja (changes requested):

web3_proxy/src/frontend/admin.rs
yenicelik: I dont think this segment changes anything with the user-tier, no?

entities/src/admin_increase_balance_receipt.rs
yenicelik: added

web3_proxy/src/frontend/admin.rs
yenicelik: modified

web3_proxy/src/frontend/admin.rs
```diff
@@ -33,6 +41,141 @@ use std::sync::Arc;
...
+        balance::Entity::insert(balance_entry)
+            .on_conflict(
+                OnConflict::new()
+                    .values([
+                        // (
+                        //     balance::Column::Id,
+                        //     Expr::col(balance::Column::Id).add(self.frontend_requests),
+                        // ),
+                        (
+                            balance::Column::AvailableBalance,
```
yenicelik, line 154: @WyseNynja is this type correct? I was looking at the docs but couldn't figure out how to do a simple update with this syntax (thus instead it's doing "insert on conflict", which will always result in a conflict)

web3_proxy/src/frontend/admin.rs
yenicelik: ok i removed this logic completely, right now there might be a "design void" with this though, admins will have to upgrade users manually to premium

web3_proxy/src/frontend/admin.rs
yenicelik: I was checking the active model functionality, couldn't find anything either. If you stumble across something please do let me know.

Issue #66, opened 2023-05-11
Add a column "out of funds" that we can check


yenicelik:
still to be discussed, but originated here: https://github.com/llamanodes/web3-proxy/pull/44#discussion_r1186571720

WyseNynja:
I think looking at `authorization.checks.current_balance < $threshold` will be simplest

yenicelik:
I added the following enum to authorization checks:
```
#[derive(Clone, Debug, Default)]
pub enum UserTier {
    #[default]
    Free,
    PrivateDemo,
    EffectivelyUnlimited,
    Unlimited,
    PremiumOutOfFunds,
    Premium,
}
```
next to balance. 

I will invalidate the item in the cache from the stats (which will trigger refetching logic for user tiers) whenever the threshold is reached.

yenicelik:
I added the "I think looking at authorization.checks.current_balance < $threshold will be simplest" logic, I will close this issue as it is outdated as of now

PR #84, opened 2023-05-24
Making Balance Changes Atomic (Read and Write)

yenicelik:
#66


web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +416,51 @@ pub async fn user_balance_post(
...
+            .filter(balance::Column::Id.eq(balance_entry_id))
+            .one(&txn)
+            .await?
+            .context("Could not find the newly inserted balance_entry, something went wrong!")?;
+
+        // Then also make the user premium if the payment is above 10$
+        if new_balance_entry.available_balance >= Decimal::new(10, 0) {
+            let mut active_recipient = recipient.clone().into_active_model();
+            // Make the recipient premium "Effectively Unlimited"
+            active_recipient.user_tier_id = sea_orm::Set(premium_user_tier.id);
```
WyseNynja, line 460: Don't actually change the row database. The row should always be their actual tier. 

Instead, the code that fetches the rate limits should check the balance. If the balance is low, use the out_of_funds_tier_id for the user. 

Changing it in the actual database will cause problems when we have many tiers

WyseNynja (changes requested):
I don't think using txn like this will work correctly. 

1) the txn needs to select rows with "FOR UPDATE" or it doesn't solve any issues with parallel requests
2) the txn needs to be committed at the end or nothing will actually save

yenicelik:
(Don't merge yet), I added `.lock(LockType::Update)` to most select queries. This, plus using transactions, should make transactions atomic. Can you validate this? If this looks alright, I will start to take care of updating the user roles within the cache.

web3_proxy/src/frontend/users/payment.rs
WyseNynja: Again, we still do not need this here. We don't want balance changes changing the column ever. Delete this. 

The way it should work, is that when a user registers, they are given the default premium role. As they won't a balance yet, the code will give them the limits of the out of funds tier that matches this role. As soon as they do have a balance, they code will see that, and they will get the limits of the premium role.

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +421,35 @@ pub async fn user_balance_post(
...
+        if user_balance.available_balance.unwrap() >= Decimal::new(10, 0) {
+            let mut active_recipient = recipient.clone().into_active_model();
+            // Make the recipient premium "Effectively Unlimited"
+            active_recipient.user_tier_id = sea_orm::Set(premium_user_tier.id);
+            active_recipient.save(&txn).await?;
+        }
+
         debug!("Setting tx_hash: {:?}", tx_hash);
         let receipt = increase_on_chain_balance_receipt::ActiveModel {
             tx_hash: sea_orm::ActiveValue::Set(hex::encode(tx_hash)),
```
WyseNynja, line 455: hex::encode will not work correctly. It does not include the 0x prefix. 

Use ether's types and their .encode_hex() functions instead. They already do the correct thing.

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +421,35 @@ pub async fn user_balance_post(
...
-
-                info!("New user balance model is: {:?}", active_user_balance);
-                active_user_balance.save(&txn).await?;
-                // txn.commit().await?;
-                // user_balance // .try_into_model().unwrap()
-            }
-        };
+        // Gotta add to the balance entry
+        // Should be atomic, because we do a write lock during the transaction
+        let mut active_user_balance = user_balance.clone().into_active_model();
```
WyseNynja, line 438: Why does this need a clone? It's probably safer to use the returned value from the save function.

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -384,10 +385,17 @@ pub async fn user_balance_post(
...
 
+        // Create a new transaction that will be used for joint transaction
+        let txn = db_conn.begin().await?;
+
+        // We must (1) lock the user and (2) lock the balance
+        // Both balance and lock must be present
+        // no need to do OnConflict with this functionality
         // Encoding is inefficient, revisit later
         let recipient = match user::Entity::find()
             .filter(user::Column::Address.eq(&recipient_account.encode()[12..]))
```
WyseNynja, line 396: Which "encode" function is this? Ethers types usually have "encode_hex" which includes the 0x prefix. If we are inconsistent about including the prefix, things will be broken.

web3_proxy/src/stats/mod.rs
```diff
@@ -222,6 +223,9 @@ impl BufferedRpcQueryStats {
             .unwrap_or(Decimal::from(0));
     }
 
+    /// Check a user's balance and possibly downgrade him in the cache
+    async fn downgrade_user(self) {}
```
WyseNynja, line 227: No. Don't do it do it like this. See my other comment

WyseNynja: Also, opening and holding this many locks across multiple users at once seems really likely to me to cause a deadlock.

if I refer you, and we both make queries at the same time on different servers, both of our stats updates will try to lock our users. And your stats update will also try to lock my user. That's a problem. 

I think we need to do the 'Expr' code like I first suggested

WyseNynja (changes requested):

web3_proxy/src/frontend/users/payment.rs
yenicelik: sorry for this! for some reason that's an artefcat that wasnt removed during the last merge ... will remove this once and for all now!

yenicelik: oh wait nevermind, this was a confusion from my side. logic makes sense now (that premium is default basically)

yenicelik: as a small heads up though, rn the code does not make the users premium by default, should that be also added?

web3_proxy/src/frontend/users/payment.rs
yenicelik: the database stores a Vec<u8>, and this is how I try to get to this, let me know if you still think it's wrong. this was working during a test that I ran, but can change it if you prefer another method

web3_proxy/src/frontend/users/payment.rs
yenicelik: I removed this completely and replaced it with an atomic "OnConflict Update" logic instead

web3_proxy/src/stats/mod.rs
yenicelik: I changed the logic to use Expr everywhere except on the referral table (because more complex logic). Let me know if this looks ok now. If the lock should still be removed, I think the only way around would be to do an SQL statement (unless sea-orm syntax allows more complex statements). let me know

WyseNynja (changes requested):

web3_proxy/src/frontend/users/payment.rs
yenicelik: we want to go from Address to Vec<u8> (specifically, [u8; 20], but as a Vec form because types must match). from_slice will not work here. I agree that there should be a cleaner way though

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -159,12 +156,12 @@ pub async fn user_balance_post(
         Ok(OpenRequestResult::Handle(handle)) => {
             debug!(
                 "Params are: {:?}",
-                &vec![format!("0x{}", hex::encode(tx_hash))]
+                &vec![format!("0x{}", tx_hash.encode_hex())]
```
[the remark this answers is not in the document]
yenicelik: good catch! i removed all occurrences, as well as the HexFmt crate

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -410,66 +413,31 @@ pub async fn user_balance_post(
...
+                    .values([(
+                        balance::Column::AvailableBalance,
+                        Expr::col(balance::Column::AvailableBalance).add(amount),
+                    )])
+                    .to_owned(),
+            )
+            .exec(&txn)
             .await?
-            .context("Could not find 'Premium' Tier in user-database")?;
+            .last_insert_id;
```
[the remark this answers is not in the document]
yenicelik: You're right, no need to get the id

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -410,66 +413,31 @@ pub async fn user_balance_post(
...
+            available_balance: sea_orm::Set(amount),
+            used_balance: sea_orm::Set(Decimal::new(0, 0)),
+            user_id: sea_orm::Set(recipient.id),
+        };
 
-        // Get the premium user-tier
-        let premium_user_tier = user_tier::Entity::find()
-            .filter(user_tier::Column::Title.eq("Premium"))
-            .one(&db_conn)
+        // TODO: Do an insert on conflict update ...
```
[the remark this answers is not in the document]
yenicelik: removed

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -410,66 +413,31 @@ pub async fn user_balance_post(
...
-                }
-
-                info!("New user balance model is: {:?}", active_user_balance);
-                active_user_balance.save(&txn).await?;
-                // txn.commit().await?;
-                // user_balance // .try_into_model().unwrap()
-            }
-        };
-        debug!("Setting tx_hash: {:?}", tx_hash);
+        debug!("Setting tx_hash: {:?}", tx_hash.encode_hex());
```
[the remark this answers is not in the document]
yenicelik: done

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -483,7 +451,7 @@ pub async fn user_balance_post(
         let response = (
             StatusCode::CREATED,
             Json(json!({
-                "tx_hash": tx_hash,
+                "tx_hash": tx_hash.encode_hex(),
```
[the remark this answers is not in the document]
yenicelik: removed

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
+            sender_bonus_applied = referral_entity.credits_applied_for_referee;
+
+            // Calculate if we are above the usage threshold, and apply a bonus
+            // Optimally we would read this from the balance, but if we do it like this, we only have to lock a single table (much safer w.r.t. deadlocks)
+            if !referral_entity.credits_applied_for_referee
+                && (referral_entity.credits_applied_for_referrer * (Decimal::from(10))
+                    + self.sum_credits_used)
+                    >= Decimal::from(100)
+            {
+                sender_available_balance_delta += Decimal::from(100);
```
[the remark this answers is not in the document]
yenicelik: This constant should probably be removed. 

A user gets 100$ free credits if they have spent 100$ (premium users only). This is the part that checks this.

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
+                sender_bonus_applied = true;
             }
 
-            // Only downgrade the user in local process memory, not elsewhere
+            // Calculate how much the referrer should get, limited to the last 12 months
+            // Apply 10% of the used balance as a bonus if applicable
+            let now = Utc::now();
+            let valid_until = DateTime::<Utc>::from_utc(referral_entity.referral_start_date, Utc)
+                .checked_add_months(Months::new(12))
+                .unwrap();
```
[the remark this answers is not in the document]
yenicelik: yh compiler doesnt complain

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
-            .await?
-        {
-            Some(x) => x,
-            None => {
-                warn!(
-                    "No referral code was found for this user: {:?}",
-                    sender_user_id
-                );
-                return Ok(());
+            if now <= valid_until {
```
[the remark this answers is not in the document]
yenicelik: so lifetime referrals? 

i agree that this can be circumvened, but i would assume just making it a bit tougher will already have large effect. I don't think everyone will do this (the majority of people wouldnt I suppose)

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
-                );
-                return Ok(());
-            }
+        // ================================= //
+        //  UPDATE REFERRER & USER BALANCE   //
+        // ================================= //
+        let user_balance = balance::ActiveModel {
+            id: sea_orm::NotSet,
+            available_balance: sea_orm::Set(sender_available_balance_delta),
+            used_balance: sea_orm::Set(sender_used_balance_delta),
```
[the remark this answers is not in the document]
yenicelik: Can we make this in a separate PR, otherwise it will make this PR super convoluted. i'll open an issue as well

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
+                .await?
+                .last_insert_id;
         }
 
-        active_sender_balance.save(db_conn).await?;
-        active_referee.save(db_conn).await?;
+        // ================================ //
+        // TODO: REFRESH USER ROLE IN CACHE //
+        // ================================ //
+        txn.commit().await?;
```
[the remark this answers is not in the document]
yenicelik: done

web3_proxy/src/stats/mod.rs
```diff
@@ -318,216 +320,177 @@ impl BufferedRpcQueryStats {
...
+                )
+                .exec(&txn)
+                .await?
+                .last_insert_id;
         }
 
-        active_sender_balance.save(db_conn).await?;
-        active_referee.save(db_conn).await?;
+        // ================================ //
+        // TODO: REFRESH USER ROLE IN CACHE //
```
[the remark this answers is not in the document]
yenicelik: This is in https://github.com/llamanodes/web3-proxy/pull/89 , I will not modify this logic here

yenicelik:
Is devel ready to be merged in? (I did it on the other PR, it looked like it's not exactly ready). Let me know if it is and I will resolve conflicts

yenicelik:
@WyseNynja can we merge this one? It would be good to merge this before #66

WyseNynja (approved):

PR #89, opened 2023-05-28
David/66 downgrade balance logic

yenicelik:
Includes part of #84

- Creates a cache for the balance with TTL 120 seconds
- Cache (also for RPC keys) is invalidated if premium is downgraded, or if payment occurs


web3_proxy/src/app/mod.rs
```diff
@@ -74,6 +76,46 @@ pub static APP_USER_AGENT: &str = concat!(
...
+            UserTier::PrivateDemo => "PrivateDemo".to_owned(),
+            UserTier::EffectivelyUnlimited => "EffectivelyUnlimited".to_owned(),
+            UserTier::Unlimited => "Unlimited".to_owned(),
+            UserTier::PremiumOutOfFunds => "PremiumOutOfFunds".to_owned(),
+            UserTier::Premium => "Premium".to_owned(),
+        }
+    }
+}
+
+impl TryFrom<&str> for UserTier {
```
WyseNynja, line 103: Why do we need this? This seems incredibly fragile to me. If we wanted user tiers in code like this, we should just remove them entirely from the database. 

There are going to be a bunch of tiers created for specific users. We don't want to have to coordinate deploys with tier changes. 

This needs to work differently

web3_proxy/src/app/mod.rs
```diff
@@ -110,7 +152,7 @@ pub struct AuthorizationChecks {
     /// IMPORTANT! Once confirmed by a miner, they will be public on the blockchain!
     pub private_txs: bool,
     pub proxy_mode: ProxyMode,
-    pub balance: Option<Decimal>,
+    pub user_tier: UserTier,
```
WyseNynja, line 155: Why is this needed? We don't generally care about having the tier. We just want the limits, and those are already part of this cached data

web3_proxy/src/app/mod.rs
```diff
@@ -128,6 +170,7 @@ impl DatabaseReplica {
 
 /// Cache data from the database about rpc keys
 pub type RpcSecretKeyCache = Arc<CacheWithTTL<RpcSecretKey, AuthorizationChecks>>;
+pub type UserBalanceCache = Arc<CacheWithTTL<u64, Decimal>>; // Could also be an AtomicDecimal
```
WyseNynja, line 173: Other places we use the user id, we use NonZeroU64. I don't know how much that matters though

web3_proxy/src/frontend/authorization.rs
```diff
@@ -432,6 +441,7 @@ impl RequestMetadata {
             response_timestamp: 0.into(),
             start_instant: Instant::now(),
             stat_sender: app.stat_sender.clone(),
+            latest_balance: latest_balance.unwrap_or(Decimal::from(0)),
```
WyseNynja, line 444: I usually use `unwrap_or_default`

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1083,6 +1093,34 @@ impl Web3ProxyApp {
         }
     }
 
+    /// Get the balance for the user.
+    ///
+    /// If a subuser calls this function, the subuser needs to have first attained the user_id that the rpc key belongs to.
+    /// This function should be called anywhere where balance is required (i.e. only rpc calls, I believe ... non-rpc calls don't really require balance)
+    pub(crate) async fn balance_checks(&self, user_id: u64) -> Web3ProxyResult<Decimal> {
+        // Downgrade the user if the balance is too low ...
```
WyseNynja, line 1101: Is this comment accurate? I don't see any downgrade logic here. It's just getting the balance, right?

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1083,6 +1093,34 @@ impl Web3ProxyApp {
...
+                .get_or_insert_async(&user_id, async move {
+                    let db_replica = self
+                        .db_replica()
+                        .web3_context("Getting database connection")?;
+
+                    let balance = balance::Entity::find()
+                        .filter(balance::Column::UserId.eq(user_id))
+                        .one(db_replica.conn())
+                        .await?
+                        .expect("related balance")
```
WyseNynja, line 1115: Do not expect. There are tons of users in the db without balance rows and having this kind of expect was killing all of their requests before. This must be `unwrap_or_default`.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1186,6 +1206,44 @@ impl Web3ProxyApp {
...
+                        let user_model = user::Entity::find_by_id(rpc_key_model.user_id)
+                            .one(db_replica.conn())
+                            .await?
+                            .expect("related user");
+
+                        let balance = balance::Entity::find()
+                            .filter(balance::Column::UserId.eq(user_model.id))
+                            .one(db_replica.conn())
+                            .await?
+                            .expect("related balance")
```
WyseNynja, line 1219: No expect!

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1186,6 +1206,44 @@ impl Web3ProxyApp {
...
+                            .available_balance;
+
+                        let mut user_tier_model =
+                            user_tier::Entity::find_by_id(user_model.user_tier_id)
+                                .one(db_replica.conn())
+                                .await?
+                                .expect("related user tier");
```
WyseNynja, line 1226: Don't expect. Use Anyhow::context and a `?`

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1186,6 +1206,44 @@ impl Web3ProxyApp {
...
+                        if user_tier_model.title == UserTier::Premium.to_string()
+                            && balance < Decimal::from(10)
+                        {
+                            // Find the equivalent downgraded user tier, and modify limits from there
+                            if let Some(downgrade_user_tier) = user_tier_model.downgrade_tier_id {
+                                user_tier_model =
+                                    user_tier::Entity::find_by_id(downgrade_user_tier)
+                                        .one(db_replica.conn())
+                                        .await?
+                                        .expect("downgrade user tier for premium");
```
WyseNynja, line 1239: Don't expect. Use anyhow context.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1186,6 +1206,44 @@ impl Web3ProxyApp {
...
+                            } else {
+                                return Web3ProxyResult::Err(Web3ProxyError::InvalidUserTier);
+                            }
+                        }
+
+                        let user_tier = UserTier::try_from(user_tier_model.title.as_str())?;
```
WyseNynja, line 1245: This won't work when we have a bunch of custom tiers. We don't need a custom model. Why not just use the actual user_tier database object?

web3_proxy/src/frontend/errors.rs
```diff
@@ -459,6 +460,17 @@ impl Web3ProxyError {
                     },
                 )
             }
+            Self::InvalidUserTier => {
+                warn!("InvalidUserTier");
+                (
+                    StatusCode::INTERNAL_SERVER_ERROR,
+                    JsonRpcErrorData {
+                        message: Cow::Borrowed("UserTier is not valid!"),
+                        code: StatusCode::BAD_REQUEST.as_u16().into(),
```
WyseNynja, line 469: Is this an internal server error, or a bad request?

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +421,26 @@ pub async fn user_balance_post(
         // Check if the item is in the database. If it is not, then add it into the database
         let user_balance = balance::Entity::find()
             .filter(balance::Column::UserId.eq(recipient.id))
-            .one(&db_conn)
-            .await?;
+            .lock(LockType::Update)
+            .one(&txn)
+            .await?
+            .context("Could not find User balance, this should have been created when signing up the user!")?;
```
WyseNynja, line 427: For new users, sure. But old users won't have a row. Just default to 0

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +421,26 @@ pub async fn user_balance_post(
...
 
         // Get the premium user-tier
         let premium_user_tier = user_tier::Entity::find()
-            .filter(user_tier::Column::Title.eq("Premium"))
-            .one(&db_conn)
+            .filter(user_tier::Column::Title.eq(UserTier::Premium.to_string()))
```
WyseNynja, line 431: This will need to change to match my comments above.

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -413,60 +421,26 @@ pub async fn user_balance_post(
...
-                active_user_balance.save(&txn).await?;
-                // txn.commit().await?;
-                // user_balance // .try_into_model().unwrap()
-            }
-        };
+        // Gotta add to the balance entry
+        // Should be atomic, because we do a write lock during the transaction
+        let mut active_user_balance = user_balance.clone().into_active_model();
+        active_user_balance.available_balance =
+            sea_orm::Set(user_balance.available_balance + amount);
```
WyseNynja, line 440: Because of how referral locking is happening, I don't think that "select for update" will work well anymore.

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -477,9 +451,30 @@ pub async fn user_balance_post(
...
+            .await?;
+
         txn.commit().await?;
         debug!("Saved to db");
 
+        app.user_balance_cache
+            .remove(&user_balance.user_id.unwrap());
+
+        for rpc_key_entity in rpc_keys {
+            // TODO: Not sure which one was inserted, just delete both ...
```
WyseNynja, line 469: The whole point of this enum is that you don't need to know which was used. They both hash the same.

`.remove(&rpc_key_entity.secret_key.into())`

web3_proxy/src/frontend/users/subuser.rs
```diff
@@ -208,7 +208,7 @@ pub async fn modify_subuser(
 
     debug!("User tier is: {:?}", user_tier);
     // TODO: This shouldn't be hardcoded. Also, it should be an enum, not sth like this ...
-    if user_tier.id != 6 {
+    if UserTier::try_from(user_tier.title.as_str())? != UserTier::Premium {
```
WyseNynja, line 211: Don't require premium to create a referral. Only require it to actually get credits for referrals

web3_proxy/src/stats/stat_buffer.rs
```diff
@@ -186,7 +187,8 @@ impl StatBuffer {
                         self.chain_id,
                         db_conn,
                         key,
-                        self.rpc_secret_key_cache.as_ref(),
+                        self.rpc_secret_key_cache.clone(),
+                        self.user_balance_cache.clone(), // Is cloning the Cache ok
```
WyseNynja, line 191: Cloning is cheap, but you can probably just pass by reference here

WyseNynja (changes requested):

WyseNynja (changes requested):
Changes needed

web3_proxy/src/app/mod.rs
yenicelik: I removed this, there are snippets like `if user_tier.title.as_str() != "Premium" {` for which I thought it would be better to have a struct. I changed it back now, let me know if that's better.

web3_proxy/src/app/mod.rs
yenicelik: removed it, was an artefact of me playing around with the cache

web3_proxy/src/app/mod.rs
yenicelik: The database saves it as a u64 that why I was using it. I will change it to `NonZeroU64`, let's keep it consistent

web3_proxy/src/frontend/authorization.rs
yenicelik: no it's not, i removed it

web3_proxy/src/frontend/authorization.rs
yenicelik: i removed this logic completely, see other comments for (possible) minor implications

web3_proxy/src/frontend/errors.rs
yenicelik: it is currently called when there is no downgrade_tier_id (while it is required), so it should be an internal server error

web3_proxy/src/frontend/users/payment.rs
yenicelik: I was thinking that we should backfill all balances to 0 when applying the migrations. I think it would easier as we remove edge-cases and keep the data more uniform. What do you think?

web3_proxy/src/frontend/users/payment.rs
yenicelik: I changed it back to the previous (red) version

web3_proxy/src/frontend/users/payment.rs
yenicelik: I removed the locking for users, only kept it for the referral table. i will test it on the referral table

web3_proxy/src/frontend/users/payment.rs
yenicelik: ok, I am only using the .into() one. Was unsure about it before

web3_proxy/src/frontend/users/subuser.rs
yenicelik: I will open a new issue for this

web3_proxy/src/stats/stat_buffer.rs
yenicelik: looks like it's got to be a clone

web3_proxy/src/bin/web3_proxy_cli/migrate_stats_to_v2.rs
```diff
@@ -169,6 +172,8 @@ impl MigrateStatsToV2 {
 
                     let request_ulid = Ulid::new();
 
+                    let latest_balance = Arc::new(AtomicF64::from(0.));
```
WyseNynja, line 175: lets just try it as a Decimal in a lock for now. If its a problem, we can come back to atomics

web3_proxy/src/frontend/admin.rs
```diff
@@ -137,29 +136,14 @@ pub async fn admin_increase_balance(
...
-                    // ),
-                    // (
-                    //     balance::Column::UserId,
-                    //     Expr::col(balance::Column::UserId).add(self.no_servers),
-                    // ),
-                ])
+                .values([(
+                    balance::Column::AvailableBalance,
+                    Expr::col(balance::Column::AvailableBalance).add(amount),
+                )])
```
WyseNynja, line 142: https://github.com/llamanodes/web3-proxy/pull/84#discussion_r1210950758

web3_proxy/src/frontend/authorization.rs
```diff
@@ -417,6 +425,11 @@ impl RequestMetadata {
             }
         }
 
+        // Fetch the balance from the cache
+        let latest_balance = match NonZeroU64::try_from(authorization.checks.user_id) {
+            Err(_) => Arc::new(AtomicF64::from(0.)),
+            Ok(x) => app.user_balance_cache.get(&x).unwrap_or_default(),
```
WyseNynja, line 431: I don't think `get` is the right thing here. if the cache entry gets evicted before this get is called, then a user with an actual balance might see `0` here. Maybe we want to call `balance_checks` here and use its return? Or maybe `get_or_insert_async` would work?

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1084,6 +1098,39 @@ impl Web3ProxyApp {
...
 
+    /// Get the balance for the user.
+    ///
+    /// If a subuser calls this function, the subuser needs to have first attained the user_id that the rpc key belongs to.
+    /// This function should be called anywhere where balance is required (i.e. only rpc calls, I believe ... non-rpc calls don't really require balance)
+    pub(crate) async fn balance_checks(&self, user_id: u64) -> Web3ProxyResult<Arc<AtomicF64>> {
+        if user_id == 0 {
+            return Ok(Arc::new(AtomicF64::from(0.)));
+        }
+        match NonZeroU64::try_from(user_id) {
```
WyseNynja, line 1109: this is redundant. the only time a NoneZeroU64 will error is if its 0. i'd delete 1106-08 and keep this match.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1084,6 +1098,39 @@ impl Web3ProxyApp {
...
+                    .get_or_insert_async(&x, async move {
+                        let db_replica = self
+                            .db_replica()
+                            .web3_context("Getting database connection")?;
+
+                        let balance: Decimal = balance::Entity::find()
+                            .filter(balance::Column::UserId.eq(user_id))
+                            .one(db_replica.conn())
+                            .await?
+                            .context("did not find related balance")?
```
WyseNynja, line 1122: if there is no related balance, default to 0.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1084,6 +1098,39 @@ impl Web3ProxyApp {
...
+                            .available_balance;
+
+                        Ok(Arc::new(AtomicF64::from(balance.to_f64().context(
+                            "Balance is too high, or too low to turn into a float!",
```
WyseNynja, line 1126: as mentioned above, lets try an `Arc<tokio::sync::RwLock<Decimal>>` for now

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1187,8 +1234,43 @@ impl Web3ProxyApp {
...
+                            .one(db_replica.conn())
+                            .await?
+                            .context("user model was not found, but every rpc_key should have a user")?;
+
+                        let balance = balance::Entity::find()
+                            .filter(balance::Column::UserId.eq(user_model.id))
+                            .one(db_replica.conn())
+                            .await?
+                            .context("related balance was not found but every user should have a balance")?
+                            .available_balance;
```
WyseNynja, line 1248: with so many legacy users, I don't think this is true. default to 0

```
.await?.map(|x| x.available_balance).unwrap_or_default();
```

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1187,8 +1234,43 @@ impl Web3ProxyApp {
...
+
+                        let mut user_tier_model =
+                            user_tier::Entity::find_by_id(user_model.user_tier_id)
+                                .one(db_replica.conn())
+                                .await?
+                                .context("related user tier not found, but every user should have a tier")?;
+
+                        // TODO: Do the logic here, as to how to treat the user, based on balance and initial check
+                        // Clear the cache (not the login!) in the stats if a tier-change happens (clear, but don't modify roles)
+                        if user_tier_model.title == "Premium" && balance < Decimal::from(10) {
```
WyseNynja, line 1258: Lets do `if user_tier_model.downgrade_tier_id.is_some() && balance < Decimal::from(0.01)` (or instead of `balance <= 0.01`, do `deposits <= spend`)

Requiring a $10 minimum deposit but then de-activating the account at $10 is probably not what we want. If a user runs a tiny bit negative, I don't think its much of a problem.

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1187,8 +1234,43 @@ impl Web3ProxyApp {
...
+                            // Find the equivalent downgraded user tier, and modify limits from there
+                            if let Some(downgrade_user_tier) = user_tier_model.downgrade_tier_id {
+                                user_tier_model =
+                                    user_tier::Entity::find_by_id(downgrade_user_tier)
+                                        .one(db_replica.conn())
+                                        .await?
+                                        .context("finding the downgrade user tier for premium did not work for premium")?;
+                            } else {
+                                return Web3ProxyResult::Err(Web3ProxyError::InvalidUserTier);
```
WyseNynja, line 1267: i don't think we want an error if this happens. they are just part of a tier that doesn't have downgrades enabled. There will definitely be accounts like that (unlimited accounts used for internal things like L2 syncing)

web3_proxy/src/frontend/errors.rs
WyseNynja: ok. the code should change from BAD_REQUEST to INTERNAL_SERVER_ERROR

But downgrade_tier_id is not required, so I think we can just delete this error

WyseNynja (changes requested):

web3_proxy/src/frontend/authorization.rs
yenicelik: done

web3_proxy/src/frontend/authorization.rs
yenicelik: done

web3_proxy/src/frontend/authorization.rs
yenicelik: It turns out i didnt need that code segment, I deleted it. may lead to merge conflicts with #84 but thats ok

web3_proxy/src/frontend/authorization.rs
yenicelik: this segment is within an if-else branch that assumes `if user_tier_model.title == "Premium" && balance < Decimal::from(10) {`. I think we do want it to error in this case

web3_proxy/src/frontend/admin.rs
yenicelik: The origin of this code snippet is from the other PR, so I will resolve this comment here

web3_proxy/src/bin/web3_proxy_cli/migrate_stats_to_v2.rs
yenicelik: ok, changing it now. Just for the record, it will be a Arc<RwLock<Decimal>>

web3_proxy/src/frontend/authorization.rs
yenicelik: implementing this

web3_proxy/src/frontend/authorization.rs
yenicelik: that makes sense, added this

web3_proxy/src/frontend/authorization.rs
yenicelik: ok I changed a bunch, specifically getting items from the cache.

web3_proxy/src/stats/mod.rs
```diff
@@ -224,26 +232,14 @@ impl BufferedRpcQueryStats {
...
+    async fn _save_db_stats(
+        &self,
         chain_id: u64,
         db_conn: &DatabaseConnection,
-        key: RpcQueryKey,
+        key: &RpcQueryKey,
         rpc_secret_key_cache: RpcSecretKeyCache,
         user_balance_cache: UserBalanceCache,
     ) -> Web3ProxyResult<()> {
-        if key.response_timestamp == 0 {
```
WyseNynja, line 233: This is a safety check, so I think we want to keep this. Though I suppose it could be a debug_assert

web3_proxy/src/stats/mod.rs
```diff
@@ -328,197 +324,335 @@ impl BufferedRpcQueryStats {
...
+                self
+            )));
+        }
+
+        // First of all, save the statistics to the database:
+        self._save_db_stats(
+            chain_id,
+            db_conn,
+            &key,
+            rpc_secret_key_cache.clone(),
```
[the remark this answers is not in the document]
yenicelik: Yes indeed, I changed it

web3_proxy/src/stats/mod.rs
```diff
@@ -328,197 +324,335 @@ impl BufferedRpcQueryStats {
...
-        *latest_balance = *latest_balance - self.sum_credits_used;
+        *latest_balance = *latest_balance + deltas.sender_available_balance_delta;
         if *latest_balance < Decimal::from(0) {
             *latest_balance = Decimal::from(0);
         }
 
         // Also check if the referrer is premium (thought above 10$ will always be treated as premium at least)
         // Should only refresh cache if the premium threshold is crossed
-        if balance_before >= self.sum_credits_used && *latest_balance < self.sum_credits_used {
+        if balance_before > Decimal::from(0) && *latest_balance == Decimal::from(0) {
```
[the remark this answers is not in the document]
yenicelik: So i think this snipper makes more sense when you check this line here: https://github.com/llamanodes/web3-proxy/pull/89/files/dd188a8473e165ca6847d17b82187ee22a58c75c..61d764545229d0f1bdba4b1065e77d0057ec6757#diff-2a86e780e3d3f3d9b9eea5a64ae983154c999af874920e336a1dace03e0b3dceR549 

Basically we remove the key from the cache when the user runs out of usable balance. 

I suppose in a next iteration we will replace this by "total_credits_used" and "total_credits_deposited", in which your logic would hold. 

In other words also, this snippet is used to determine whether the cache should be cleared (to make the tier an OutOfFunds tier, for example)

Let me know if you see anything off with this

web3_proxy/src/stats/mod.rs
yenicelik: just to confirm, you want to put this into _save_db_stats()?

Right now, my code shows me this: 
```
async fn save_db(
        self,
        chain_id: u64,
        db_conn: &DatabaseConnection,
        key: RpcQueryKey,
        rpc_secret_key_cache: &RpcSecretKeyCache,
        user_balance_cache: &UserBalanceCache,
    ) -> Web3ProxyResult<()> {
        if key.response_timestamp == 0 {
            return Err(Web3ProxyError::Anyhow(anyhow!(
                "no response_timestamp! This is a bug! {:?} {:?}",
                key,
                self
            )));
        }

        // First of all, save the statistics to the database:
        self._save_db_stats(
            chain_id,
            db_conn,
            &key,
            rpc_secret_key_cache,
            user_balance_cache,
        )
        .await?;
        ```

Issue #85, opened 2023-05-24
Remove llamanodes- prefix for referral


yenicelik:

yenicelik:
PR opened will close this https://github.com/llamanodes/web3-proxy/pull/90

PR #90, opened 2023-05-29
referral code does not have llamanodes- prefix any longer

yenicelik:
Removed llamanodes- prefix from referral code

#85

### DIFF
--- a/web3_proxy/src/referral_code.rs
+++ b/web3_proxy/src/referral_code.rs
@@ -8,7 +8,7 @@ pub struct ReferralCode(String);
 impl Default for ReferralCode {
     fn default() -> Self {
         let out = Ulid::new();
-        Self(format!("llamanodes-{}", out))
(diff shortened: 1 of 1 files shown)


web3_proxy/src/referral_code.rs
```diff
@@ -22,11 +22,13 @@ impl TryFrom<String> for ReferralCode {
     type Error = anyhow::Error;
 
     fn try_from(x: String) -> Result<Self> {
-        if !x.starts_with("llamanodes-") {
-            return Err(anyhow::anyhow!(
-                "Referral Code does not have the right format"
-            ));
-        }
+        // TODO: Check if string is a valid ULID
```
WyseNynja, line 25: i don't think we actually want this todo. it can be useful to make "vanity" referral codes that won't have the default ULID structure.

WyseNynja (approved):

Issue #65, opened 2023-05-11
Fix subuser balance logic


yenicelik:
https://github.com/llamanodes/web3-proxy/pull/44#discussion_r1186554879

https://github.com/llamanodes/web3-proxy/pull/44#discussion_r1186610966

https://github.com/llamanodes/web3-proxy/pull/44#discussion_r1186874565

https://github.com/llamanodes/web3-proxy/pull/44#discussion_r1186874764

yenicelik:
- [ ] Secondary users should be able to see revert logs, too
- [ ] Should we also fetch secondary keys here? (this is displayed in a separate endpoint for clear separation)
- [ ] Possible problem here. A secondary user isn't required to have a balance of their own. We are planning to disable viewing dashboards if their balance is zero though. So I think we need to also have an entry for secondary balances. Maybe a hashmap. That way we can allow a secondary user access to dashboards.

yenicelik:
@WyseNynja not sure if I understand the last point and where the hashmap comes into play

yenicelik:
also reg. point 2 and 3, those are displayed on a separate endpoint `/subuser/rpc_keys`

yenicelik:
@WyseNynja i'm also confused about the last comment here, would appreciate any more info!

WyseNynja:
So dashboard stats should not be allowed until the user has paid us $10, but secondary user’s complicate this logic.

I think some examples will demonstrate it best

# Example 1:
- User A is the owner of Key A
- User A has a balance of $100
- User B is a collaborator for Key A
- User B has no balance

User A should be able to see their own stats
User B should be able to see stats of Key A.
User B should **not** be able to see their own stats.

# Example 2:
- User A is the owner of Key A
- User A has an balance
- User B is a collaborator for Key A
- User B has no balance

Neither User A nor B should be able to see stats

yenicelik:
Closed in last PR

WyseNynja:
theres still some edge cases that don't work

WyseNynja:
I think this is fixed now

PR #96, opened 2023-05-29
Missing subuser logic

yenicelik:
https://github.com/llamanodes/web3-proxy/issues/65


web3_proxy/src/frontend/users/stats.rs
```diff
@@ -53,8 +53,22 @@ pub async fn user_revert_logs_get(
         .await
         .web3_context("failed loading user's key")?;
 
+    // Also add rpc keys for which this user has access
+    let shared_rpc_keys = secondary_user::Entity::find()
+        .filter(secondary_user::Column::UserId.eq(user.id))
+        .all(db_replica.conn())
+        .await?
+        .into_iter()
+        .map(|x| x.rpc_secret_key_id);
```
WyseNynja, line 62: the frontend team asked that we also keep track of the secondary_user's permissions. that way they can show/hide various elements in the UI.

web3_proxy/src/frontend/users/stats.rs
```diff
@@ -53,8 +53,22 @@ pub async fn user_revert_logs_get(
...
+
+    // We shouldn't need to be deduped, bcs the set of shared keys is distinct from the user's keys,
+    // the database also handles deduplication bcs it's a projection operation
     // TODO: only select the ids
-    let uks: Vec<_> = uks.into_iter().map(|x| x.id).collect();
+    let uks: Vec<_> = uks
+        .into_iter()
+        .map(|x| x.id)
```
WyseNynja, line 69: this should attach a "owner" role to this id. that way it can be zipped up with the shared_rpc_keys

web3_proxy/src/frontend/users/stats.rs
yenicelik: we do keep track of secondary user permissions, i will filter the map accordingly

web3_proxy/src/frontend/users/stats.rs
```diff
@@ -53,13 +56,42 @@ pub async fn user_revert_logs_get(
...
+    #[derive(Serialize)]
+    struct OutTuple {
+        id: u64,
+        role: Role,
+    };
+
+    // Also add rpc keys for which this user has access
+    let shared_rpc_keys = secondary_user::Entity::find()
+        .filter(secondary_user::Column::UserId.eq(user.id))
+        .all(db_replica.conn())
```
WyseNynja, line 68: with #99 merged, `db_replica.conn` -> `db_replica.as_ref`

WyseNynja (changes requested):
tiny fix needed

web3_proxy/src/frontend/users/stats.rs
yenicelik: merged from devel

WyseNynja (approved):

Issue #101, opened 2023-05-31
Change available_balance and used_balance to total_deposits and total_cost


yenicelik:
Instead of available_balance and used_balance, lets change this to total_deposits and total_cost. I think that will be a lot simpler to use and update.

PR #109, opened 2023-06-07
Change balance to record total spend and total deposits 

yenicelik:
#101


web3_proxy/src/stats/mod.rs
```diff
@@ -370,14 +371,16 @@ impl BufferedRpcQueryStats {
...
         let mut deltas = Deltas {
-            sender_available_balance_delta: -self.sum_credits_used,
-            sender_used_balance_delta: self.sum_credits_used,
+            balance_used_outside_free_tier: Default::default(),
+            balance_used_including_free_tier: Default::default(),
             sender_bonus_applied: false,
-            referrer_available_balance_delta: Decimal::from(0),
+            referrer_deposit_delta: Default::default(),
+            sender_bonus_balance_deposited: Default::default(),
         };
```
WyseNynja, line 384: nit: Deltas could derive(Default) and then this would just be `Deltas::default()`

WyseNynja (approved):

web3_proxy/src/stats/mod.rs
yenicelik: good point

PR #120, opened 2023-06-11
added some caching for influx queries, really fast provided same parameters

yenicelik:
@WyseNynja If the frontend makes the same query the cache should kick in. Specifically, if they round the intervals to the nearest 5 minutes (for example), caches should work well. I didn't constraint the cache further (division into daily, monthly, etc.), as this gives the frontend more flexibility to work with. 

Let me know if this is alongst the lines of what you imagined.

Optimally all parameters are set (no "defaults" are used, including query_start and query_stop), and any timestamps are clipped to the nearest 5-minute interval (depending on granularity)

### DIFF
--- a/web3_proxy/src/frontend/mod.rs
+++ b/web3_proxy/src/frontend/mod.rs
@@ -11,10 +11,12 @@ pub mod status;
 pub mod users;
 
 use crate::app::Web3ProxyApp;
+use axum::body::Bytes;
 use axum::{
     routing::{get, post, put},
     Extension, Router,
 };
+use hashbrown::HashMap;
 use http::{header::AUTHORIZATION, StatusCode};
 use listenfd::ListenFd;
 use log::{debug, info};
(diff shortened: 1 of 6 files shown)


WyseNynja:
a few comments from slack:

- [ ] is adding sha256 sums like that okay? seems like it should be xor. But I think we should just have the cache key be the actual params. I'm not sure hashing like this will actually be fastest
- [ ] the cache should work more like the other cache in the frontend
    - the value i think should be a Bytes. I don't think we will need the status code as part of the cached value since we only need to cache 200s. serde has a lot of overhead so this should keep memory down
    - and the cache should have a weigh function that uses the Bytes' `.len()`

yenicelik:
Implemented up to the Bytes part that you mentioned, I couldn't figure out how to go from the json! object to the response-bytes value. Please let me know if you know how

web3_proxy/src/frontend/mod.rs
```diff
@@ -58,6 +63,14 @@ pub async fn serve(
         .time_to_live(Duration::from_secs(1))
         .build();
 
+    // Move this to the mod.rs (frontend) server, this does not have to live inside the app
+    // These responses are pretty large, so we will only save up to 1000 at a time. They can live for longer, however
+    let influx_cache: InfluxResponseCache = CacheBuilder::new(1_000_000)
```
yenicelik, line 68: @WyseNynja capacity is very high, but having it less made the cache not work out

WyseNynja:
i'm not even sure we need this. the global stats load very fast now

WyseNynja:
if query costs in influx end up being high, we can look into it. but i'd rather keep our RAM for jsonrpc caches

yenicelik:
I agree with this. Also keeping code leaner is probably better

PR #129, opened 2023-06-16
Handle uncle transactions

WyseNynja:


WyseNynja:
need to update tests to handle the changes to the disconnect logic

PR #140, opened 2023-06-22
stripe integration first mvp

yenicelik:


entities/src/stripe_increase_balance_receipt.rs
```diff
@@ -0,0 +1,40 @@
+//! `SeaORM` Entity. Generated by sea-orm-codegen 0.10.6
+
+use sea_orm::entity::prelude::*;
+use serde::Serialize;
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq, Serialize)]
+#[sea_orm(table_name = "stripe_increase_balance_receipt")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
```
WyseNynja, line 10: I32 should be u64

migration/src/m20230622_104142_stripe_deposits.rs
```diff
@@ -0,0 +1,104 @@
...
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        manager
+            .create_table(
+                Table::create()
+                    .table(StripeIncreaseBalanceReceipt::Table)
+                    .if_not_exists()
```
WyseNynja, line 14: Remove the "not_exists" checks. We do our locking such that all steps should always succeed

migration/src/m20230622_104142_stripe_deposits.rs
```diff
@@ -0,0 +1,104 @@
...
+                    )
+                    .col(
+                        ColumnDef::new(StripeIncreaseBalanceReceipt::StripePaymentIntendId)
+                            .string()
+                            .not_null(),
+                    )
+                    .col(
+                        // I will not mark this as "not null", so we can track who to refund more easily if needed
+                        ColumnDef::new(StripeIncreaseBalanceReceipt::DepositToUserId)
+                            .big_unsigned(),
```
WyseNynja, line 30: Unsigned

web3_proxy/src/frontend/mod.rs
```diff
@@ -169,6 +169,14 @@ pub async fn serve(
         .route("/user", post(users::user_post))
         .route("/user/balance", get(users::payment::user_balance_get))
         .route("/user/deposits", get(users::payment::user_deposits_get))
+        .route(
+            "/user/stripe/deposits",
```
WyseNynja, line 173: Should we put the get and the post on the same route?

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+    // Return the response, all except the user ...
+    let response = json!({
+        "user": Address::from_slice(&user.address),
+        "deposits": receipts,
+    });
+
+    Ok(Json(response).into_response())
+}
+
+/// `POST /user/balance/stripe/` -- Process a confirmed txid to update a user's balance.
```
WyseNynja, line 64: Comment should be about stripe, not txids

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+#[debug_handler]
+pub async fn user_balance_stripe_post(
+    Extension(app): Extension<Arc<Web3ProxyApp>>,
+    ip: Option<InsecureClientIp>,
```
WyseNynja, line 68: Ip shouldn't be optional. That was me trying to debug. It will always be set

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+        return Err(Web3ProxyError::AccessDenied);
+    };
+
+    // Get the payload, and the header
+    let payload = params.remove("data").ok_or(Web3ProxyError::BadRequest(
+        "You have not provided a 'data' for the Stripe payload".into(),
+    ))?;
+
+    // TODO Get this from the header
+    let signature = params
```
WyseNynja, line 94: There is a "HeaderMap" extractor for this

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+        .remove("STRIPE_SIGNATURE")
+        .ok_or(Web3ProxyError::BadRequest(
+            "You have not provided a 'STRIPE_SIGNATURE' for the Stripe payload".into(),
+        ))?;
+
+    // Now parse the payload and signature
+    // TODO: Move env variable elsewhere
```
WyseNynja, line 101: Add the api config as an Option<String> to AppConfig. If not set, give a 500 error here.

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+    debug!("Found PaymentIntent Event: {:?}", intent);
+
+    if intent.status.to_string() != "succeeded" {
+        return Ok("Received Webhook".into_response());
+    }
+
+    let db_conn = app.db_conn().context("query_user_stats needs a db")?;
+
+    if stripe_increase_balance_receipt::Entity::find()
+        .filter(stripe_increase_balance_receipt::Column::Id.eq(intent.id.as_str()))
```
WyseNynja, line 126: Wrong column. Should be StripePaymentIntendId

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+    let txn = db_conn.begin().await?;
+
+    // Assert that it's usd
+    if intent.currency.to_string() != "USD" || !recipient.is_some() {
+        // In this case I should probably still save it to the database,
+        // but not increase balance (this should be refunded)
+        // TODO: I suppose we could send a refund request right away from here
+        error!(
+            "Please refund this transaction! {} {}",
+            intent.currency, &recipient_email
```
WyseNynja, line 179: Include the intent Id in this log

WyseNynja (changes requested):

web3_proxy/Cargo.toml
```diff
@@ -89,12 +89,13 @@ serde = { version = "1.0.164", features = [] }
...
 serde_prometheus = "0.2.2"
 siwe = "0.5.2"
+async-stripe = { version = "0.22.2", features = ["runtime-tokio-hyper-rustls"] }
 strum = { version = "0.25.0", features = ["derive"] }
-time = "0.3.22"
 tokio = { version = "1.28.2", features = ["full"] }
 tokio-console = { version = "0.1.8", optional = true }
 tokio-stream = { version = "0.1.14", features = ["sync"] }
 tokio-uring = { version = "0.4.0", optional = true }
+time = "0.3"
```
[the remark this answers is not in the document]
WyseNynja: actually, why is this added? i think chrono is in the migration prelude

entities/src/stripe_increase_balance_receipt.rs
yenicelik: fixed

migration/src/m20230622_104142_stripe_deposits.rs
yenicelik: fixed

migration/src/m20230622_104142_stripe_deposits.rs
```diff
@@ -0,0 +1,104 @@
...
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        manager
+            .create_table(
+                Table::create()
+                    .table(StripeIncreaseBalanceReceipt::Table)
+                    .if_not_exists()
+                    .col(
+                        ColumnDef::new(StripeIncreaseBalanceReceipt::Id)
+                            .big_integer()
```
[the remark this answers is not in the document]
yenicelik: fixed

migration/src/m20230622_104142_stripe_deposits.rs
yenicelik: this should be big-unsigned, no? otherwise it complains on my end about wrong foreign keys types

web3_proxy/Cargo.toml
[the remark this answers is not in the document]
yenicelik: the async-stripe package has a bunch of dependencies; we were using time=0.3.3 before, but it was incompatible, so i relaxed versioning to 0.3 . alphabetized as well by now

web3_proxy/src/frontend/mod.rs
yenicelik: hmm good point actually, let me change that later; works for me, also decreases naming complexity

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: ok fixed

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+        "Could not parse the stripe webhook request!".into(),
+    ))?;
+
+    let intent = match event.data.object {
+        stripe::EventObject::PaymentIntent(intent) => intent,
+        _ => return Ok("Received irrelevant webhook".into_response()),
+    };
+    debug!("Found PaymentIntent Event: {:?}", intent);
+
+    if intent.status.to_string() != "succeeded" {
```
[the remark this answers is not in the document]
yenicelik: yes changed it

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+        .await?
+        .is_some()
+    {
+        return Ok("Payment was already recorded".into_response());
+    };
+
+    let recipient_email = match intent.receipt_email {
+        Some(x) => x,
+        None => {
+            error!("Please check this! This payment did not provide any e-mail");
```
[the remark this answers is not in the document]
yenicelik: the next log logs the user_id that was passed as a parameter

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    let recipient = user::Entity::find()
+        .filter(user::Column::Email.eq(&recipient_email))
+        .one(&db_conn)
+        .await?;
+
+    // Filter for the intent.id one, and return false if it already exists
+
+    // we do a fixed 2 decimal points because we only accept USD for now
+    let amount = Decimal::new(intent.amount, 2);
```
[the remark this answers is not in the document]
yenicelik: Yes, stripe returns integers with the lowest cent-amount: https://stripe.com/docs/currencies#zero-decimal
Decimal::new( . , 2) makes the last two digits acknowledge as decimal digits

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,240 @@
...
+    let recipient_id = match &recipient {
```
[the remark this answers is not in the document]
yenicelik: done

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: added

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: fixed

web3_proxy/src/frontend/mod.rs
yenicelik: changed

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: changed

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: added, this is sensitive so should never be versioned an exposed

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: added

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,236 @@
...
+    };
+
+    let recipient_user_id: u64 = params
+        .remove("user_id")
+        .ok_or(Web3ProxyError::BadRouting)?
+        .parse()
+        .or(Err(Web3ProxyError::ParseAddressError))?;
+
+    // Get the payload, and the header
+    let payload = params.remove("data").ok_or(Web3ProxyError::BadRequest(
```
WyseNynja, line 95: Is the payload actually in the headers? Isn't it in the POST body?

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,236 @@
...
+        "Could not parse the stripe webhook request!".into(),
+    ))?;
+
+    let intent = match event.data.object {
+        stripe::EventObject::PaymentIntent(intent) => intent,
+        _ => return Ok("Received irrelevant webhook".into_response()),
+    };
+    debug!("Found PaymentIntent Event: {:?}", intent);
+
+    if intent.status.as_str() != "succeeded" {
```
WyseNynja, line 129: I feel like we probably want to do something more if it's not succeeded. Like, what if they post to tell us there was a chargeback?

web3_proxy/src/frontend/users/payment_stripe.rs
```diff
@@ -0,0 +1,236 @@
...
+        )
+        .one(&db_conn)
+        .await?
+        .is_some()
+    {
+        return Ok("Payment was already recorded".into_response());
+    };
+
+    let recipient: Option<user::Model> = user::Entity::find()
+        .filter(user::Column::Id.eq(recipient_user_id))
```
WyseNynja, line 147: You can use find_by_id here

web3_proxy/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![feature(build_hasher_simple_hash_one)]
```
WyseNynja, line 1: What requires this?

WyseNynja (changes requested):

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: according to this example, data is in payload, and the signature is in the header: https://stripe.com/docs/webhooks

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: Screenshot 2023-06-26 at 20 03 18

which ones do you want to cover? also as I understand, this endpoint responds to the webhook, not the frontend, so not sure how we would best communicate it with the frontend

web3_proxy/src/lib.rs
yenicelik: async-stripe requires it, there's many dependencies. lmk if you want to remove this, it would require a deeper rework of the stripe api though (definitely would take 1-2 full days to separate the parts we're interested in)

web3_proxy/src/frontend/users/payment_stripe.rs
WyseNynja: Ok. This is wrong then. This needs to use the `Json` body extractor instead of the Params extractor

web3_proxy/src/frontend/users/payment_stripe.rs
WyseNynja: What do we need to communicate to the frontend? They can just poll the user's balance endpoint, right?

web3_proxy/src/lib.rs
WyseNynja: I'm fine with it. Maybe just add a comment saying what it's for (since it's a lib and not our own code like the other features)

web3_proxy/src/frontend/users/payment_stripe.rs
yenicelik: yes agree. i'll leave it at this then

WyseNynja (approved):
Since it looks like the stripe library does their own json parsing, I think serde_json::RawValue might be better than Value. Perf doesn't matter too much on this, but it will save us deserializing it to Value, serializing it to String, and then the stripe library deserializing it again.

WyseNynja:
to deal with the conflict on the time package, maybe switch to using chrono. We already have it in other places in the code.

WyseNynja:
replaced by #154

PR #156, opened 2023-07-03
Update Dockerfile

sambacha:


Dockerfile
```diff
@@ -22,7 +22,9 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
     libsasl2-dev \
     libzstd-dev \
     make \
-    pkg-config
+    pkg-config;
+    apt-get clean; \
+	rm -rf /var/lib/apt/lists/*;
```
WyseNynja, line 27: There's no need to run clean or clear apt lists since they are part of a mount type cache. That will just make the next build take longer

WyseNynja (changes requested):

Dockerfile
```diff
@@ -7,11 +7,11 @@ ENV PATH "/root/.foundry/bin:/root/.cargo/bin:${PATH}"
 
 # install rustup dependencies
 # also install web3-proxy system dependencies. most things are rust-only, but not everything
-RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+RUN set -eux; \
```
WyseNynja, line 10: why change to `set -e` but also keep the `&&`? Those can change to `;`

WyseNynja:
I changed things to use set -eux so that logs are a bit easier to follow. didn't change the --cache though

PR #158, opened 2023-07-05
Split errors

WyseNynja:
- [ ] 0 cost errors


entities/src/login.rs
```diff
@@ -9,8 +8,7 @@ use serde::{Deserialize, Serialize};
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: u64,
-    #[sea_orm(unique)]
-    #[serde(serialize_with = "serialization::uuid_as_ulid")]
```
yenicelik, line 13: we don't need the serialize decorator anymore?

entities/src/pending_login.rs
```diff
@@ -9,8 +8,7 @@ use serde::{Deserialize, Serialize};
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: u64,
-    #[sea_orm(unique)]
-    #[serde(serialize_with = "serialization::uuid_as_ulid")]
```
yenicelik, line 13: same question as before

entities/src/revert_log.rs
```diff
@@ -13,7 +12,7 @@ pub struct Model {
     pub rpc_key_id: u64,
     pub timestamp: DateTimeUtc,
     pub method: Method,
-    #[serde(serialize_with = "serialization::vec_as_address")]
```
yenicelik, line 16: Same question as before. Also, is Vec<u8> intended? before, it was marked as a uuid

entities/src/rpc_key.rs
```diff
@@ -10,8 +10,7 @@ pub struct Model {
     #[sea_orm(primary_key)]
     pub id: u64,
     pub user_id: u64,
-    #[sea_orm(unique)]
-    #[serde(serialize_with = "serialization::uuid_as_ulid")]
+    #[sea_orm(column_type = "Binary(BlobSize::Blob(Some(16)))", unique)]
```
yenicelik, line 13: same question as before

entities/src/user.rs
```diff
@@ -9,8 +8,7 @@ use serde::{Deserialize, Serialize};
 pub struct Model {
     #[sea_orm(primary_key)]
     pub id: u64,
-    #[sea_orm(unique)]
-    #[serde(serialize_with = "serialization::vec_as_address")]
+    #[sea_orm(column_type = "Binary(BlobSize::Blob(Some(20)))", unique)]
```
yenicelik, line 11: again a Vec<u8>, no need to change with Address? also is the serialization decorator intended

web3_proxy/src/stats/influxdb_queries.rs
```diff
@@ -575,7 +576,25 @@ pub async fn query_user_stats<'a>(
                             );
                         }
                         _ => {
-                            error!("error_response should always be a Long!");
+                            error!("error_response should always be a String!");
+                        }
+                    }
+                } else if stat_response_type == StatType::Detailed && key == "user_error_response" {
```
yenicelik, line 582: just confirming that this was added to the influx schema

yenicelik (approved):
Looks good, a couple questions just to make sure

entities/src/login.rs
WyseNynja: we do. just missed it being replaced

web3_proxy/src/stats/influxdb_queries.rs
WyseNynja: Don't we use a schemaless influx that gets set on first insert?

yenicelik: i suppose we do

PR #159, opened 2023-07-06
David/test referral

WyseNynja:


web3_proxy/tests/common/get_rpc_key.rs
```diff
@@ -0,0 +1,58 @@
...
+use web3_proxy::frontend::admin::AdminIncreaseBalancePost;
+use web3_proxy::frontend::users::authentication::{LoginPostResponse, PostLogin};
+
+#[derive(Debug, Deserialize)]
+pub struct RpcKeyResponse {
+    pub user_id: u64,
+    pub user_rpc_keys: std::collections::HashMap<String, RpcKey>,
+}
+
+#[derive(Debug, Deserialize)]
```
WyseNynja, line 17: can we re-use an existing type, or do we not have this exact layout already somewhere else?

web3_proxy/tests/common/get_rpc_key.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub allowed_ips: Option<serde_json::Value>,
+    pub allowed_origins: Option<serde_json::Value>,
+    pub allowed_referers: Option<serde_json::Value>,
+    pub allowed_user_agents: Option<serde_json::Value>,
+    pub description: Option<serde_json::Value>,
+    pub id: u64,
+    pub log_revert_chance: f64,
+    pub private_txs: bool,
+    pub role: String,
+    pub secret_key: String,
```
WyseNynja, line 29: should we have a stricter type on this? (ulid or RpcSecretKey)

web3_proxy/tests/test_users.rs
```diff
@@ -73,11 +88,136 @@ async fn test_log_in_and_out() {
...
+            .clone()
+            .unwrap()
+    );
+
+    // Make a for-loop just spam it a bit
+    // Make a JSON request
+    // TODO: Also get the RPC key for the user first ...
+    let rpc_keys: RpcKey = user_get_first_rpc_key(&x, &r, &user_login_response).await;
+    info!("Rpc key is: {:?}", rpc_keys);
+    info!(?rpc_keys);
```
WyseNynja, line 178: why log twice?

yenicelik: no need

web3_proxy/tests/common/get_rpc_key.rs
yenicelik: I thought about keeping it String just because that's what the client will interpret it as. can change it to Ulid though.

web3_proxy/tests/common/get_rpc_key.rs
yenicelik: we could, i just did it for convenience right now. i'll check which ones we can re-use

web3_proxy/tests/common/get_rpc_key.rs
WyseNynja: Lets do Ulid. There was one time I messed up and was returning them as Uuid by accident. This test could catch that

yenicelik:
outdated

PR #165, opened 2023-07-06
Lock type update instead of exclusive

WyseNynja:


web3_proxy/src/frontend/users/payment.rs
```diff
@@ -236,7 +237,7 @@ pub async fn user_balance_post(
 
     // check for uncles
     let mut find_uncles = increase_on_chain_balance_receipt::Entity::find()
-        // .lock_exclusive()
```
WyseNynja, line 239: i'm not sure if we want this, but it seems like it should have a lock

yenicelik: Sounds about right.

PR #172, opened 2023-07-09
David/referral more tests

WyseNynja:


migration/src/m20230708_151756_rpc_accounting_free_usage_credits.rs
```diff
@@ -0,0 +1,40 @@
...
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(RpcAccountingV2::Table)
+                    .add_column(
+                        ColumnDef::new(RpcAccountingV2::SumInclFreeCreditsUsed)
+                            .decimal_len(20, 10)
+                            .not_null(),
```
WyseNynja, line 16: should this have `.default(0)`?

migration/src/m20230708_152131_referral_track_one_time_bonus_bonus.rs
```diff
@@ -0,0 +1,48 @@
...
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        manager
+            .alter_table(
+                Table::alter()
+                    .table(Referee::Table)
+                    .drop_column(Referee::CreditsAppliedForReferee)
+                    .add_column(
+                        ColumnDef::new(Referee::OneTimeBonusAppliedForReferee)
+                            .decimal_len(20, 10)
+                            .not_null(),
```
WyseNynja, line 17: as above, should this have `.default(0)`?

migration/src/m20230708_225009_balance_view_v2.rs
```diff
@@ -0,0 +1,48 @@
+use sea_orm_migration::prelude::*;
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
+
+#[async_trait::async_trait]
+impl MigrationTrait for Migration {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        // Replace the sample below with your own migration scripts
+        todo!();
```
WyseNynja, line 10: https://discord.com/channels/873880840487206962/900758302294704188/1127456562865258548

from my searching, it looks like sea_orm doesn't have any official support for Views. This means their tooling won't create the entities for us or help us write the migrations.

Given that it is a read only table and we will just be doing a simple `.filter(User::Id.eq(user_id)` on it, i think we should just add the query select logic to the existing `Balance` model. Given the view isn't a "materialized view", I wonder if we should even bother. We can just write the sql in a  `async fn for_user(id: u64) -> Balance`

migration/src/m20230708_151756_rpc_accounting_free_usage_credits.rs
yenicelik: another inconsistency i found in the table; sometimes its marked as .default("0.0"), .default(0.0), and one time not at all

migration/src/m20230708_225009_balance_view_v2.rs
yenicelik: where would you want to introduce this function?

PR #173, opened 2023-07-09
David/remove balance logic non destructive

WyseNynja:


web3_proxy/src/frontend/authorization.rs
```diff
@@ -1308,11 +1262,11 @@ impl Web3ProxyApp {
                         // TODO: Do the logic here, as to how to treat the user, based on balance and initial check
                         // Clear the cache (not the login!) in the stats if a tier-change happens (clear, but don't modify roles)
                         if let Some(downgrade_user_tier) = user_tier_model.downgrade_tier_id {
-                            let balance = latest_balance.read().clone();
+                            let balance = latest_balance.read().await.clone();
 
                             // only consider the user premium if they have paid at least $10 and have a balance > $.01
                             // otherwise, set user_tier_model to the downograded tier
-                            if balance.total_deposit < Decimal::from(10)
+                            if balance.total_deposits > Decimal::from(10)
```
WyseNynja, line 1269: Why flip this? I'm pretty sure it was correct. We want to use the downgrade tier id in two cases. 

1. They haven't yet paid $10
2. They have paid >$10 but their balance is < $0.01

web3_proxy/src/frontend/users/payment.rs
```diff
@@ -436,7 +413,7 @@ pub async fn user_balance_post(
             match NonZeroU64::try_from(recipient.id) {
                 Err(_) => {}
                 Ok(x) => {
-                    app.user_balance_cache.invalidate(&x).await;
+                    app.user_balance_cache.invalidate(&x.get()).await;
```
WyseNynja, line 416: What does .get do here?

WyseNynja (commented):
`get_balance_from_db` should probably be a function on Balance named `try_from_db`

web3_proxy/src/frontend/users/payment.rs
yenicelik: it gets u64 from NonZeroU64. Unwrapping NonZeroU64 make the code more confuse, so I decided to change the UserBalanceCache to Cache<u64, Balance>, instead of Cache<NonZeroU64, Balance>. Most times user_id is a u64, so it's easier to deal with. Lmk if you think this is an important blocker (i.e. should be changed back to NonZeroU64)

web3_proxy/src/frontend/authorization.rs
yenicelik: ah ok that#s what it is, sorry for some reason I read the || as an &&, thus the confusion. Thanks! I changed it back

web3_proxy/src/frontend/authorization.rs
```diff
@@ -1147,60 +1135,26 @@ impl Web3ProxyApp {
...
-                    }
-                })
-                .await
-                .map_err(Into::into),
-        }
+    ) -> Web3ProxyResult<Arc<AsyncRwLock<Balance>>> {
+        self.user_balance_cache
+            .try_get_with(user_id, async move {
+                let db_replica = self.db_replica()?;
+                loop {
```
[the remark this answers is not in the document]
yenicelik: the initial place where I copy-pastd it from had a loop (I suppose for retrying if there was a database error). i suppose this is not required, the user would just make a second request in this very edge case.

web3_proxy/src/config.rs
```diff
@@ -6,6 +6,7 @@ use ethers::prelude::{Address, TxHash};
 use ethers::types::{U256, U64};
 use hashbrown::HashMap;
 use migration::sea_orm::DatabaseConnection;
+use rust_decimal::Decimal;
```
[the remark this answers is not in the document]
yenicelik: Does stripe have features we can turn on or off? As I understand this is not even all the features. Can you verify that it's just "core" features when no features are set? (see API Features section)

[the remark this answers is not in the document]
WyseNynja: I turned off all the stripe features we could. Some of the ones we needed are really broad

web3_proxy/src/frontend/users/payment.rs
WyseNynja: It's fine. The types are just getting in the way most of the time. I wish converting was less verbose

PR #177, opened 2023-07-10
split balance queries up

WyseNynja:


web3_proxy/src/balance.rs
```diff
@@ -1,80 +1,166 @@
...
-            DbBackend::MySql,
-            raw_sql,
-            // [.into()],
-        ))
-        .one(db_conn)
-        .await?
-        {
-            None => return Ok(None),
-            Some(x) => x,
+        let (admin_deposits,) = admin_increase_balance_receipt::Entity::find()
```
yenicelik, line 79: only issue i see here is that the query may not be atomic. i dont know if that will also lead to race conditions.

web3_proxy/src/balance.rs
```diff
@@ -1,80 +1,166 @@
...
     pub fn remaining(&self) -> Decimal {
-        self.total_deposits - self.total_spent_paid_credits
+        self.total_deposits() - self.total_spent_paid_credits
+    }
+
+    pub fn total_deposits(&self) -> Decimal {
+        self.admin_deposits
+            + self.chain_deposits
+            + self.referal_bonus
+            + self.referee_bonus
```
yenicelik, line 68: maybe rename to one-time-bonus or so, otherwise its very easy to mix this one up with referral bonus

yenicelik: renamed

PR #197, opened 2023-07-27
auto-generate entities. this breaks things. manual patching required

BlinkyStitt:
this currently fails to compile because we changed the foreign key index on rpc_accounting_v2 to just be a regular index. The SQL query still works. sea-orm just doesn't see it as a related entitity anymore and so `.inner_join(rpc_key::Entity)` fails.

I'm not sure the recommended way to fix this. For now we can just stay on the old entities. If we have to migrate before fixing this, the relationship can be put back after the entities are generated.


WyseNynja:
after reviewing this, i see that it didn't actually make any changes that we want. I'm going to save undoing these changes (ignoring the version numbers) as a patch. Then we can apply the patch after generating and not have to clean up anything by hand again.

PR #212, opened 2023-09-14
streaming responses

WyseNynja:


web3_proxy/src/rpcs/one.rs
```diff
@@ -1020,7 +1030,7 @@ impl Hash for Web3Rpc {
         self.name.hash(state);
 
         // TODO: url does NOT include the authorization data. i think created_at should protect us if auth changes without anything else
```
WyseNynja, line 1032: remove this todo. the url has the auth data in it now (which is good imo)

web3_proxy/src/rpcs/one.rs
```diff
@@ -173,6 +178,7 @@ impl Web3Rpc {
         } else {
             None
         };
+        */
```
WyseNynja, line 181: lets just delete this now

web3_proxy/src/rpcs/request.rs
```diff
@@ -194,15 +195,35 @@ impl OpenRequestHandle {
...
+                    .rpc
+                    .next_id
+                    .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+                let payload = JsonRpcRequest::new(
+                    JsonRpcId::Number(id),
+                    method.to_string(),
+                    serde_json::to_value(params)?,
+                )
+                .unwrap();
+                // TODO no early return
```
WyseNynja, line 210: why no early return? what do you mean by this?

web3_proxy/src/rpcs/request.rs
```diff
@@ -194,15 +195,35 @@ impl OpenRequestHandle {
...
+                let resp = client
+                    .post(self.rpc.http_url.clone().expect("must have url"))
```
WyseNynja, line 212: can we prepare the client with this url already included in it? seems like we shouldn't need this string every time

web3_proxy/src/rpcs/request.rs
```diff
@@ -194,15 +195,35 @@ impl OpenRequestHandle {
...
+                    .json(&payload)
+                    .send()
+                    .await?;
+                // TODO configurable size
+                StreamingResponse::read_if_short(resp, 1024).await
+            } else if let Some(p) = self.rpc.ws_provider.load().as_ref() {
+                p.request(method, params)
+                    .await
+                    .map(StreamingResponse::Buffer)
```
WyseNynja, line 221: love it

WyseNynja (commented):
im really curious to see what this does for perf

PR #233, opened 2023-11-07
forward request id for eth_sendRawTransaction

talking-llama:
currently looking into a single test case failure which may be due to this change.


web3_proxy/src/jsonrpc/request_builder.rs
```diff
@@ -394,11 +400,16 @@ impl ValidatedRequest {
...
         #[cfg(feature = "rdkafka")]
         let kafka_debug_logger = if matches!(authorization.checks.proxy_mode, ProxyMode::Debug) {
-            // TODO: get this out of tracing instead (where we have a String from Amazon's LB)
-            let request_ulid = Ulid::new();
+            let request_ulid = request_id
+                .map(|s| Ulid::from_string(&s))
+                .transpose()
+                .ok()
+                .flatten()
+                .unwrap_or_else(Ulid::new);
```
WyseNynja, line 413: don't force it to a Ulid. That should be the fallback. we want the amazon header

WyseNynja (changes requested):

